PR #3, opened 2021-04-20
support for endport

haim-kermany:

PR #22, opened 2021-04-28
policy type checking

shireenf-ibm:


network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -36,12 +36,16 @@ def parse_label_selector_requirement(self, requirement, namespace_selector):
         """
         self.check_keys_are_legal(requirement, 'requirement', {'key': 1, 'operator': 1, 'values': 0})
```
shai-doron, line 37: Add type check in function `check_keys_are_legal`

shai-doron (commented):
good job so far. let's finish the work on this item

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -298,25 +296,21 @@ def parse_policy(self):
...
-        if not isinstance(policy_metadata, dict):
-            self.syntax_error('type of metadata is not a map in NetworkPolicy', self.policy)
         allowed_metadata_keys = {'name': 1, 'namespace': 0, 'annotations': 0, 'clusterName': 0, 'creationTimestamp': 0,
                                  'deletionGracePeriodSeconds': 0, 'deletionTimestamp': 0, 'finalizers': 0,
                                  'generateName': 0, 'generation': 0, 'labels': 0, 'managedFields': 0,
                                  'ownerReferences': 0, 'resourceVersion': 0, 'selfLink': 0, 'uid': 0}
         self.check_keys_are_legal(policy_metadata, 'network policy metadata', allowed_metadata_keys)
         if not isinstance(policy_metadata['name'], str):
-            self.syntax_error('The name field is not a string in the NetworkPolicy metadata', policy_metadata['name'])
+            self.syntax_error('type of name is not string in metadata of NetworkPolicy', policy_metadata['name'])
```
shai-doron, line 305: type of name is not `a` string

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -298,25 +296,21 @@ def parse_policy(self):
...
         if 'namespace' in policy_metadata:
             if not isinstance(policy_metadata['namespace'], str):
-                self.syntax_error('The namespace field is not a string in the NetworkPolicy metadata',
+                self.syntax_error('type of namespace is not string in metadata of NetworkPolicy',
```
shai-doron, line 308: type of namespace is not `a` string

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -326,7 +320,7 @@ def parse_policy(self):
             self.warning('policyTypes is missing/empty in the spec of ' + res_policy.full_name(), policy_spec)
         else:
             if not isinstance(policy_types, list):
-                self.syntax_error('policyTypes is not a string array in the spec of '
+                self.syntax_error('type of policyTypes is not a string array in spec of '
```
shai-doron, line 323: remove `string`

shai-doron (commented):
Good job @shireenf-ibm.
For keys that do not have strict allowed values to check against, we can still consider checking the type of the value. Pls consider.

shireenf-ibm:
Hi , in my last commit , i made changes following running tests on the live k8s cluster.
one of the new checks added is the metadata.name regex check , names have to match k8s pattern.
I made changes to all tests in the workspace.
but the "run all tests" will keep failing in withIpBlock2-scheme.yaml until changes be done in the master 
in this scheme there is a networkpolicylist that refers to:
- https://github.com/IBM/network-config-analyzer/blob/master/tests/example_policies/withIpBlock2/withIpBlock2-networkpolicy5.yaml

in that yaml , metadata.names does not match the regex pattern of k8s and I couldn't find a way to correct them on my branch.
attaching the fixed file (replacing B with b in "ibBlock" in name [fields)
[fixed1_ghe.txt](https://github.com/IBM/network-config-analyzer/files/6501044/fixed1_ghe.txt)

network-config-analyzer/NetworkPolicy.py
```diff
@@ -101,6 +101,9 @@ def get_policy_type(policy):
             return NetworkPolicy.PolicyType.Unknown
 
         policy_type = NetworkPolicy.PolicyType.Unknown
+        if not isinstance(kind, str):
+            return policy_type
```
shai-doron, line 105: replace lines 103-105 with the following code to be more consistent with code at lines 95-101

if not isinstance(kind, str):
   return NetworkPolicy.PolicyType.Unknown

policy_type = NetworkPolicy.PolicyType.Unknow

network-config-analyzer/GenericYamlParser.py
```diff
@@ -46,30 +46,45 @@ def warning(self, msg, obj=None):
...
+                code = key_info
+                value_type = None
             if code == 1 and key not in dict_to_check:
-                self.syntax_error(f'{dict_name} must have a {key} entry', dict_to_check)
-            if code == 2 and key in dict_to_check:
-                self.syntax_error(f'{key} is not yet supported inside {dict_name}', dict_to_check)
-            if key in dict_to_check and key in allowed_values:
+                self.syntax_error(f'{dict_name} must have {key} entry', dict_to_check)
+            if key in dict_to_check:
+                if code == 2:
```
shai-doron, line 73: can these 2 lines be unified `if code == 2 and key in dict_to_check:`

network-config-analyzer/GenericYamlParser.py
```diff
@@ -46,30 +46,45 @@ def warning(self, msg, obj=None):
...
+                    self.syntax_error(f'{key} is not yet supported inside {dict_name}', dict_to_check)
                 value = dict_to_check.get(key)
-                if value not in allowed_values[key]:
-                    self.syntax_error(f'{key} has invalid value/type in {dict_name}')
+                if code == 1 and value is None:
+                    self.syntax_error(f'mandatory {key} entry can not be null in {dict_name}', dict_to_check)
```
shai-doron, line 77: I think this is more accurate: `entry` --> `value`

network-config-analyzer/GenericYamlParser.py
```diff
@@ -46,30 +46,45 @@ def warning(self, msg, obj=None):
...
+                if value_type is not None:
+                    if value == 0 and not value_type == int:
+                        self.syntax_error(f'type of {key} is not {value_type} in {dict_name}',
+                                          dict_to_check)
+                    if value is not None and not isinstance(value, value_type):
+                        self.syntax_error(f'type of {key} is not {value_type} in {dict_name}',
+                                          dict_to_check)
+                if key in allowed_values:
```
shai-doron, line 85: I'm not sure, but isn't this should be `if key in dict_to_check and key in allowed_values:`

shai-doron (commented):
Also, I didn't see code that calls `check_keys_are_legal()` and passes `allowed_kyes` with  value_type.
I would expect to see such code for fields that their value type can be checked?

network-config-analyzer/GenericYamlParser.py
shireenf-ibm: all the "if" statements in lines 73 - 85 are indented under the "if key in dict_to_check"
otherwise I'll need to unify them all with it

network-config-analyzer/GenericYamlParser.py
shireenf-ibm: as mentioned above, this if statement is included under if key in dict_to_check

network-config-analyzer/GenericYamlParser.py
```diff
@@ -46,30 +46,45 @@ def warning(self, msg, obj=None):
...
+                    self.syntax_error(f'{key} is not yet supported inside {dict_name}', dict_to_check)
                 value = dict_to_check.get(key)
-                if value not in allowed_values[key]:
-                    self.syntax_error(f'{key} has invalid value/type in {dict_name}')
+                if code == 1 and value is None:
+                    self.syntax_error(f'mandatory {key} entry can not be null in {dict_name}', dict_to_check)
+                if value_type is not None:
+                    if value == 0 and not value_type == int:
+                        self.syntax_error(f'type of {key} is not {value_type} in {dict_name}',
+                                          dict_to_check)
```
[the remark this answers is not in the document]
shireenf-ibm: no.
for all "non integer" types python "sees" 0 as an empty value. (for example, if we have dict = 0 python translates it to {} )
empty value != none 
however , k8s always treats 0 as an integer.
So, I had to add this check "as to force python to treat 0 as int" in order to follow k8s messages

shireenf-ibm:
> 
> 
> Also, I didn't see code that calls `check_keys_are_legal()` and passes `allowed_kyes` with value_type.
> I would expect to see such code for fields that their value type can be checked?

code calls are in the k8sPolicyYamlParser.py
this file was committed with all of the changes (it appears under Files changed tab above)
mainly:
for fields that we want to check their value type , I send the type with the code in this pattern {key: [code, value_type] , ...}
for other fields , I kept the old pattern , {key: code , ...}
a check for what info we have for each key is in the beginning of the 'check_legal_keys' method in the generic parser file

network-config-analyzer/GenericYamlParser.py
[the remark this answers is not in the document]
shai-doron: ok

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -307,17 +320,19 @@ def parse_policy(self):
...
+        if ingress_rules:
+            for ingress_rule in ingress_rules:
+                res_policy.add_ingress_rule(self.parse_ingress_rule(ingress_rule, res_policy.selected_peers))
 
-        for egress_rule in policy_spec.get('egress', []):
-            res_policy.add_egress_rule(self.parse_egress_rule(egress_rule))
+        egress_rules = policy_spec.get('egress', [])
+        if egress_rules:
+            for egress_rule in egress_rules:
+                res_policy.add_egress_rule(self.parse_egress_rule(egress_rule))
```
shai-doron, line 335: can lines 328 and 333 can be omitted?
if no ingress/egress rules are specified then anyhow policy_spec.get('ingress', []) and policy_spec.get('egress', []) will return null and the code inside the following for loops to add the rule will not be executed.

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -266,39 +273,45 @@ def parse_policy(self):
...
-        res_policy = K8sNetworkPolicy(self.policy['metadata']['name'], self.namespace)
+
+        policy_metadata = self.policy['metadata']
+        allowed_metadata_keys = {'name': [1, str], 'namespace': [0, str], 'annotations': 0, 'clusterName': 0,
+                                 'creationTimestamp': 0, 'deletionGracePeriodSeconds': 0, 'deletionTimestamp': 0,
+                                 'finalizers': 0, 'generateName': 0, 'generation': 0, 'labels': 0, 'managedFields': 0,
+                                 'ownerReferences': 0, 'resourceVersion': 0, 'selfLink': 0, 'uid': 0}
+        self.check_keys_are_legal(policy_metadata, 'metadata', allowed_metadata_keys)
+        pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+        if re.fullmatch(pattern, policy_metadata['name']) is None:
```
shai-doron, line 294: I'm not sure I fully understand this. Let's discuss

shai-doron (commented):
good job

network-config-analyzer/K8sPolicyYamlParser.py
shireenf-ibm: No, then we will get error message when having the following line in yaml file (copying from the test file that caused to add these lines):
 #3 empty (spare)
  egress:

if the lines (328, 333) are omitted
python reads this "value" as None , and None != [] , so it will try to iterate over it and print this error
TypeError: 'NoneType' object is not iterable
however k8s passes with no errors in this case
so I had to add the if statements in order to follow k8s behavior

shai-doron: very good thanks for clarifying the reason

network-config-analyzer/K8sPolicyYamlParser.py
shireenf-ibm: Copying an error message that I got when running tests on live k8s cluster (in my test : name included **_** ):

> The NetworkPolicy "metadata_name_test7" is invalid: metadata.name: Invalid value: "metadata_name_test7": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

after consulting with Ziv , we want to follow k8s behavior here too, so I used the same k8s regex to our checks.

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -290,13 +318,9 @@ def parse_policy(self):
...
         self.check_keys_are_legal(policy_metadata, 'metadata', allowed_metadata_keys)
-        if len(policy_metadata['name']) > 253:
-            self.syntax_error('a metadata name must contain no more than 253 characters', self.policy)
-        pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
-        if re.fullmatch(pattern, policy_metadata['name']) is None:
-            self.syntax_error('a metadata name must consist of lower case alphanumeric characters, '
-                              '"-" or ".", and must start and end with an alphanumeric character', self.policy)
+        self.check_dns_subdomain_name('metadata name', policy_metadata['name'])
         if 'namespace' in policy_metadata and policy_metadata['namespace'] is not None:
+            self.check_dns_label_names('metadata namespace', policy_metadata['namespace'])
```
shai-doron: I believe this is wrong. name of `namespace` should comply with dns subdomain name and not dns label name.
We should call `check_dns_label_names()` on fields of `matchLabels` and `matchExpressions`

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -27,6 +27,34 @@ def __init__(self, policy, peer_container, policy_file_name=''):
...
+        """
+        checking validity of the resource name
+        :param key: The key name
+        :param value : The value assigned for the key
+        :return: None
+        """
+        if len(value) > 253:
+            self.syntax_error(f'a {key} value must be no more than 253 characters', self.policy)
+        pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+        if re.fullmatch(pattern, value) is None:
```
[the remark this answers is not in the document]
shireenf-ibm: from python docs:
 
> re.fullmatch(pattern, string, flags=0)
> 
>     If the whole string matches the regular expression pattern, return a corresponding match object. Return None if the string does not match the pattern; note that this is different from a zero-length match.

that is why I check if it is None

network-config-analyzer/K8sPolicyYamlParser.py
shireenf-ibm: from the k8s reference (namespace) : 
`


namespace string | Namespace  defines the space within which each name must be unique. An empty  namespace is equivalent to the "default" namespace, but "default" is the  canonical representation. Not all objects are required to be scoped to a  namespace - the value of this field for those objects will be empty.   **Must be a DNS_LABEL.**
-- | --




`
the metadata name is a " subdomain dns label" 

matchLabels and "key" field in matchExpressions : follows the "k8s labels and selectors reference (neither dns subdomain/ dns label)" 
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
 
Additionally, all of them were tested on k8s live cluster (on the updated excel file (attached to the issue), I added descriptions of the bad path tests and and the k8s' error messages I got)

network-config-analyzer/K8sPolicyYamlParser.py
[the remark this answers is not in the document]
shai-doron: you are right

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -27,6 +27,34 @@ def __init__(self, policy, peer_container, policy_file_name=''):
...
+        """
+        checking validity of the label values
+        :param key: The key name
+        :param value : The value assigned for the key
+        :return: None
+        """
+        if len(value) > 63:
+            self.syntax_error(f'a {key} value must be no more than 63 characters', self.policy)
+        pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+        if re.fullmatch(pattern, value) is None:
```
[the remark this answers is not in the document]
shai-doron: you are right

network-config-analyzer/K8sPolicyYamlParser.py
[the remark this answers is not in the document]
shai-doron: right

shai-doron (approved):
good job

PR #25, opened 2021-05-10
Bug fix for reading pod list from a live cluster

shai-doron:


zivnevo (changes requested):
I prefer there only one `for` loop over list items, so please merge the loop in lines 354--356 with the loop in lines 363--369.

zivnevo (approved):
ok now

Issue #37, opened 2021-05-31
crash when scanning a yaml with a list


zivnevo:
This happens when a yaml that contains a list is encountered when scanning for namespace-defining yaml.
e.g., if encountering a yaml with the following content
```
- name: no_ftp
  description: Forbid ftp into any pod
  protocol: TCP
  port_min: 20
  port_max: 21
  action: deny
- name: no_telnet
  description: Forbid telnet into any pod
  protocol: TCP
  port_min: 23
  port_max: 23
  action: deny
```

PR #38, opened 2021-05-31
Issue #37: fixing a crash when reading a yaml with a top-level list

zivnevo:


network-config-analyzer/PeerContainer.py
```diff
@@ -105,7 +105,7 @@ def _set_namespace_list(self, ns_resources):
                 with open(path) as yaml_file:
                     code = yaml.load_all(yaml_file, Loader=yaml.SafeLoader)
                     for ns_code in code:
-                        if ns_code and ns_code.get('kind') == 'NamespaceList':
+                        if isinstance(ns_code, dict) and ns_code.get('kind') == 'NamespaceList':
```
shai-doron, line 108: Shouldn't we fix as well at line 172: `if resource and resource.get('kind') == 'NamespaceList':` ?

zivnevo: good catch. I found a few more

shai-doron (dismissed):

zivnevo:
@shai-doron  - there were some issues with my previous solution. Please review again

shai-doron (approved):

PR #42, opened 2021-06-16
short switches for selected command-line settings

### LINKED ISSUE #41: Provide shorthands for command-line switches
We want to have shorthands for common command-line switches. e.g., `-s` for `--scheme` and `-n` for `--ns_list`.
The new `--o` and `--f` flags should therefore be `--output_format` and `--file_out` and should have a short version of `-o` and `-f`.

zivnevo:
Fixes #41 

Also: refactoring ExecutionAssist to contain only a single class.


network-config-analyzer/CLExecute.py
```diff
@@ -0,0 +1,173 @@
...
+from PeerContainer import PeerContainer
+from NetworkConfig import NetworkConfig
+import NetworkConfigQuery
+
+
+class CLExecute:
+    """
+    A class for running queries from commandline
+    """
+    def __init__(self, ns_list='', pod_list='', output_format='txt', output_path=None):
```
adisos, line 19: When we do not specify the flag output_format in the cli args, it sends a None value, and the default 'txt' value is not applied. So currently the OutputConfiguration object is created with outputFormat = None and we get the error : 
"unexpected outputFormat in output configuration value [should be txt or yaml],  value is: None".

adisos (changes requested):

network-config-analyzer/CLExecute.py
zivnevo: set a default value `txt` for the `--output` arg.

adisos (approved):
ok to merge

PR #48, opened 2021-06-22
Fw rules csv

adisos:


zivnevo (changes requested):
- [ ] Make the query string shorter for command-line runs as we discussed.
- [ ] Add some tests with output comparisons for csv format.

zivnevo (approved):
looks great :)

PR #76, opened 2021-08-04
fix istio authorization policy parsing + add tests

adisos:


network-config-analyzer/IstioPolicyYamlParser.py
```diff
@@ -247,6 +260,53 @@ def parse_source(self, source_dict):
...
+        """
+        key_array = dict_elem.get(key)
+        if key_array is not None:
+            self.validate_array_not_empty(key_array, key)
+            return key_array
+        return None
+
+    def validate_dict_elem_has_non_empty_array_value(self, dict_elem, dict_key_str):
+        """
+        assuming that dict values are of type arrays, checking that at least one of the arrays is not empty,
```
zivnevo, line 300: All these function should better move to `GenericYamlParser.py` - other parsers can use them in the future.

zivnevo (changes requested):

network-config-analyzer/IstioPolicyYamlParser.py
adisos: done

zivnevo (approved):

PR #83, opened 2021-08-22
Nca docker image

tanyaveksler:
Added a workflow for building and pushing docker image. 
Implemented version handling through manually maintaining source file build_number.txt.


zivnevo (changes requested):
As we discussed, we would like to have two workflows:
### Automatic docker build (maybe add to the main testing script?)
Triggered every pull request and every push to master (like all testing workflows)
* Builds a docker image
* Runs the above docker image using `docker run <image name> -h` just to make sure it is functional
### Manual docker build, push and create a release
* Creates and pushes a new git tag (e.g., `git tag -a v1.2.0 -m "Version 1.2.0" ; git push origin v1.2.0`)
* Creates a new release, using https://github.com/actions/create-release
* Builds a docker image
* Pushes built image to registry

I have second thoughts about maintaining the version in a repository file. It complicates things and adds redundant commits. How about using [workflow inputs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) to set the version manually when we run the manual flow? Seems much simpler to me.

tanyaveksler:
Regarding your thoughts about a version number: I can implement it as you suggested, using workflow input, and it indeed seems simpler. On the other hand, this way we cannot automatically increment patch number and minor version number - we should do it manually, which is also a bit complicated and human error prone. What do you prefer?

tanyaveksler:
Implemented automatic docker build and check inside test-push workflow.
Changed manual docker build and push workflow (now called make-github-and-docker-release.yml) to receive a version number parameter and to build docker release and github release.

.github/workflows/make-github-and-docker-release.yml
```diff
@@ -0,0 +1,53 @@
+name: Create docker release and github release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version number'
+        required: true
+        default: 0.0.0
```
zivnevo, line 9: Is there a meaning to `default` when `required` is set to `true`?

zivnevo (approved):
Looks good

.github/workflows/make-github-and-docker-release.yml
tanyaveksler: It's a number that appears in the box where we can enter the release number

PR #87, opened 2021-09-12
IPv6 testing

tanyaveksler:
Added policies for testing IPv6 blocks; added queries with these policies to withIpBlock2-scheme.yaml.
Added tests for handling badly-written CIDRs to network-policy-chech-bad-path-scheme.yaml.
Improved handling exception of bad CIDR (throwing syntax_error, like for any other error in a policy).
Separated printing query output to different lines, for different query parts.


network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -422,14 +422,25 @@ def parse_policy(self):
         res_policy.selected_peers &= self.peer_container.get_namespace_pods(self.namespace)
 
         ingress_rules = policy_spec.get('ingress', [])
+        exception = ''
         if ingress_rules:
             for ingress_rule in ingress_rules:
-                res_policy.add_ingress_rule(self.parse_ingress_rule(ingress_rule, res_policy.selected_peers))
+                try:
+                    res_policy.add_ingress_rule(self.parse_ingress_rule(ingress_rule, res_policy.selected_peers))
```
zivnevo, line 429: Better have this try-catch block inside `parse_ip_block`, so you don't have to repeat it for ingress and for egress.

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -422,14 +422,25 @@ def parse_policy(self):
...
+                except ValueError as e:
+                    exception = e.args
+                if exception:
+                    self.syntax_error(exception)
```
zivnevo, line 433:
```suggestion
                except ValueError as e:
                    self.syntax_error(e.args)
```

network-config-analyzer/SchemeRunner.py
```diff
@@ -228,6 +228,7 @@ def _run_equivalence(self, configs_array, output_configuration):
                 total_res += not full_result.bool_result
                 if not full_result.bool_result:
                     query_output += full_result.output_explanation + '\n'
+#                query_output += '\n'
```
zivnevo, line 231: Please clean commented-out code

tests/example_policies/withIpBlock2/withIpBlock2-scheme.yaml
```diff
@@ -32,16 +37,25 @@ networkConfigList:
...
+      - withIpBlock2-networkpolicy5-IPv6.yaml
+    expectedWarnings: 0
+
 queries:
   - name: no_emptiness
     emptiness:
       - np1
       - np2
+      - np2-ipv6
+      - np5-ipv6
```
zivnevo, line 52: Please add an `expected` value for this query

tests/example_policies/withIpBlock2/withIpBlock2-scheme.yaml
```diff
@@ -32,16 +37,25 @@ networkConfigList:
...
 
   - name: no_vacuity
     vacuity:
       - np1
       - np2
+      - np2-ipv6
+      - np5-ipv6
```
zivnevo, line 58: Add `expected` value

zivnevo (changes requested):

network-config-analyzer/K8sPolicyYamlParser.py
tanyaveksler: Initially I implemented it like you suggested, and then I saw in the output two exceptions, and the message that exception was thrown while handling another exception. I found the solution for this - added 'from None' when raising syntax_error exception.

network-config-analyzer/K8sPolicyYamlParser.py
tanyaveksler: Done

network-config-analyzer/SchemeRunner.py
tanyaveksler: Done

tests/example_policies/withIpBlock2/withIpBlock2-scheme.yaml
tanyaveksler: Done

tests/example_policies/withIpBlock2/withIpBlock2-scheme.yaml
tanyaveksler: Done

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -422,6 +425,7 @@ def parse_policy(self):
         res_policy.selected_peers &= self.peer_container.get_namespace_pods(self.namespace)
 
         ingress_rules = policy_spec.get('ingress', [])
+        exception = ''
```
zivnevo, line 428:
```suggestion
```

network-config-analyzer/K8sPolicyYamlParser.py
```diff
@@ -200,7 +200,10 @@ def parse_ip_block(self, block):
         """
         self.check_fields_validity(block, 'ipBlock', {'cidr': [1, str], 'except': [0, list]})
         res = Peer.PeerSet()
-        res.add(Peer.IpBlock(block['cidr'], block.get('except')))
+        try:
+            res.add(Peer.IpBlock(block['cidr'], block.get('except')))
+        except ValueError as e:
+            self.syntax_error(e.args)
```
zivnevo, line 206:
```suggestion
            self.syntax_error(e.args, block)
```
This should provide a context for the error (file and line number)

zivnevo (changes requested):
More things to consider:
* What happens if a policy contains both IPv4 and IPv6 addresses (in different rules or in the same rule)?
* What happens if we compare a policy with IPv4 block to a policy with IPv6 blocks? If I understand correctly, there is no overlap in the address space defined by IPv4 and IPv6.

tanyaveksler:
Added mixed Ipv4-Ipv4 policies, rules and queries.
Added handling TypeError in query running and policy parsing.
Added support of expectedError key to queries, similarly to its support in networkConfigList. This enabled adding mixed ipv4-ipv6 queries to withIpBlock2-scheme.yaml for testing.

zivnevo (changes requested):
We are going to get mixed policies in real life and we should be able to compare them without getting query errors. In general, we don't want queries to return errors at all.

To avoid the current exception, I suggest to define IPv6 addresses as always larger than IPv4 addresses, and this should allow comparing and manipulating them.
To implement that, I suggest to define in `Peer.py` a new proxy class that only holds a single network address. It also has a < operator (`__lt__()` method)  that can compare network addresses of different versions, as defined above. You may also have to define a == operator (`__eq__` method) for this class.
Then, in `IPBlock.add_cidr()` use your class to define the boundaries of the intervals and the holes.

tanyaveksler:
Implemented mixed IPv4-IPv6 comparisons by adding proxy class IPNetworkAddress.

Issue #96, opened 2021-11-01
Allow json files as netpol inputs - not just yamls


zivnevo:
This is relevant when scanning Git repos or file-system directories

PR #98, opened 2021-11-02
issue_96 supporting json as netpol inputs, code changes and tests

shireenf-ibm:
#96 
2 tests included:
1. adding json np from file entry -> under testcase1
2. adding json np from fs dir -> the dir is tests/json_dir
    the scheme file -> testcase10-json-version


zivnevo (changes requested):
Looks good.
Please put `json_dir` under `example_policies/testcase10-json-version`.
We already have too many subdirs under `tests`.

zivnevo (approved):
ok to merge

Issue #97, opened 2021-11-01
Allow multiple files for reading pod_lists


zivnevo:
Allow this in both scheme files and command-line mode.
Command-line switches `--pod_list` and `--base_pod_list` should have *append* action, so they can appear several times in the command line, and their result is a list of files to interpret as lists of pods.

PR #100, opened 2021-11-03
supporting multiple pod_resources files 

shireenf-ibm:
#97

### DIFF
--- a/network-config-analyzer/PeerContainer.py
+++ b/network-config-analyzer/PeerContainer.py
@@ -156,7 +156,7 @@ class PeerContainer:
         """
 
         if not peer_resources_list:
-           self.load_peer_from_k8s_live_cluster()
+            self.load_peer_from_k8s_live_cluster()
 
         for peer_resources in peer_resources_list:
             # load from git
--- a/network-config-analyzer/PeerContainer.py
+++ b/network-config-analyzer/PeerContainer.py
@@ -65,6 +65,17 @@ class PeerContainer:
                 return
         raise Exception('Failed to locate Kubernetes configuration files')
 
+    def load_ns_from_live_cluster(self):
+        self.locate_kube_config_file()
+        yaml_file = CmdlineRunner.get_k8s_resources('namespace')
+        ns_code = yaml.load(yaml_file, Loader=yaml.SafeLoader)
+        self.set_namespaces(ns_code)
+
+    def load_peer_from_k8s_live_cluster(self):
+        self.locate_kube_config_file()
+        peer_code = yaml.load(CmdlineRunner.get_k8s_resources('pod'), Loader=yaml.SafeLoader)
+        self.add_eps_from_list(peer_code)
+
     def _add_namespace(self, ns_object, generic_list):
         """
         Adds a single namespace to the container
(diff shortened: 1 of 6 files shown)


shireenf-ibm:
Hi @zivnevo ,
in this first commit , I've started to add the support of multiple pod list resources in the scheme runner only (cmd line support to be committed later).
I have some questions before finalizing the code:
1. in this commit I suppose that pod list urls/files are seperated by ',' (_treated as python str_ ).
Is this good? we can also write them **as python list** (with '-' before each resource), then I'll need to make this change for all the existing tests.

2. Shouldn't we do the same (supporting multiple files) to the namespace list
for example when I add this :
image
I get the following warning message (because the ns file not added):
image

zivnevo:
1. I prefer a yaml list with '-'. There is no need to change existing tests. Just check in the code for the type of the value, convert strings to a list of length 1, then iterate the list.
2. Good idea. Let's do the same for namespaceList to be consistent.
3. Ignore this warning. It's expected.

network-config-analyzer/SchemeRunner.py
```diff
@@ -56,6 +56,15 @@ def _get_input_file(self, given_path):
...
         return given_path
 
+    def _handle_resorces_list(self, resources_list):
+        if not resources_list: # shouldn't get here
+            return None
+        if isinstance(resources_list, str):
+            resources_list = [resources_list]
+        for idx, resource in enumerate(resources_list):
+            resources_list[idx] = self._get_input_file(resource)
+        return resources_list
```
zivnevo, line 66:
```suggestion
        input_file_list = []
        for resource in resources_list:
            input_file_list.append(self._get_input_file(resource))
        return input_file_list
```
Better not override the list given as input

network-config-analyzer/SchemeRunner.py
```diff
@@ -56,6 +56,15 @@ def _get_input_file(self, given_path):
             return input_file
         return given_path
 
+    def _handle_resorces_list(self, resources_list):
```
zivnevo, line 59: fix typo in function name

network-config-analyzer/PeerContainer.py
```diff
@@ -94,85 +105,91 @@ def set_namespaces(self, ns_list):
...
-                        for ep_list in peer_code:
-                            self.add_eps_from_list(ep_list)
-
-        # load from local directory
-        elif os.path.isdir(peer_resources):
-            for path in Path(peer_resources).glob('**/*.yaml'):
-                with open(path) as yaml_file:
+
+        if not peer_resources_list:
+           self.load_peer_from_k8s_live_cluster()
```
zivnevo, line 159:
```suggestion
        if not peer_resources_list:
            self.load_peer_from_k8s_live_cluster()
```
Fix indentation

network-config-analyzer/PeerContainer.py
```diff
@@ -94,85 +105,91 @@ def set_namespaces(self, ns_list):
...
+                            self.add_eps_from_list(peer_code)
+
+            # load from live cluster
+            elif peer_resources == 'calico':
+                for peer_type in ['wep', 'hep', 'networkset', 'globalnetworkset']:
+                    peer_code = yaml.load(CmdlineRunner.get_calico_resources(peer_type), Loader=yaml.SafeLoader)
+                    self.add_eps_from_list(peer_code)
+            elif peer_resources == 'k8s':
+                self.load_peer_from_k8s_live_cluster()
+
```
zivnevo, line 192: remove redundant line
```suggestion
```

zivnevo (changes requested):
Looks good. Just a few minor fixes

zivnevo (approved):
ok to merge

Issue #99, opened 2021-11-02
A new class for scanning a directory for yaml/json files


zivnevo:
Currently scanning a directory for yaml/json files has three different implementations: in `NetworkConfig::add_policies_from_fs_dir()`, in `PeerContainer::_set_namespace_list()` and in `PeerContainer::_set_peer_list()`.
It is proposed to have one class, named `DirScanner`, to provide directory-scanning utilities, pretty much in the spirit of the `GitScanner` class (authored by Shai).
Also, see if `GitScanner` and `DirScanner` share similar code (e.g., identifying a file as json/yaml) and move this code into a common base class.

PR #102, opened 2021-11-14
Issue 99 dir scanner

shireenf-ibm:
#99


network-config-analyzer/GenericScanner.py
```diff
@@ -0,0 +1,60 @@
...
+    @staticmethod
+    def is_yaml_file(file_name):
+        """
+        returns if the given file is a yaml file
+        :param str file_name: the name of the file
+        :rtype: bool
+        """
+        if file_name.endswith('.yaml') or file_name.endswith('.yml') or file_name.endswith('json'):
+            return True
+        return False
```
zivnevo, line 45:
```suggestion
        extension = os.path.splitext(file_name)[1]
        return extension in {'.yaml', '.yml', '.json'}
```

zivnevo (changes requested):
As we discussed, we want to remove duplicate code even further:
1. Use `DirScanner` for handling both directories and single files
2. Add a factory-like method to `GenericScanner`, called `get_scanner()` that will take a look at the entry and return either a `GitScanner` or a `DirScanner`. The two subclasses should then implement the method `get_yamls()`. This way the caller can simply call `GenericScanner::get_scanner().get_yamls()` and iterate over the result.

Also, please consider renaming `GenericScanner` to `GenericTreeScanner` or to any other less generic name :)

shireenf-ibm:
some comments on the newest commit:
1. the base generic class became abstract class and I've put the  factory method in a "factory class" otherwise we'll have circular dependencies. However, since I kept the factory class in the same file, I had to put its import lines not in top of file
2. in the generic scanner class replaced the py_yaml (import yaml) with ruamel.yaml because the py_yaml fails to load claico policies. In addition it fails to load calico policies in a "safe" load type , then I added a flag to indicate if we are scanning a policy (np_scanner)
3. in PeerContainer.py::(set_peer_list) method, I call only the add_eps_from_list method after loading the yaml files, because I found that it already handles all the cases that were written former for github, file and dir

network-config-analyzer/DirScanner.py
```diff
@@ -0,0 +1,36 @@
...
+        if GenericTreeScanner.is_yaml_file(file_path):
+            yield from self._yield_yaml_file(file_path, open(file_path))
+
+    def get_yamls(self):
+        """
+        Call this function to get a generator for all yaml files
+        """
+        if os.path.isfile(self.fs_path):
+            yield from self.check_and_yield_file(self.fs_path)
+
```
zivnevo, line 33:
```suggestion
            yield from self.check_and_yield_file(self.fs_path)
            return

```
We don't want to continue into the rest of the function after the first yield.

network-config-analyzer/GenericTreeScanner.py
```diff
@@ -0,0 +1,95 @@
...
+@dataclass
+class YamlFile:
+    """
+    A class for holding the retrieved data of a yaml file from git repo
+    """
+    data: object
+    path: str
+
+
+class GenericTreeScanner(metaclass=abc.ABCMeta):
```
zivnevo, line 23:
```suggestion
class GenericTreeScanner(abc.ABC):
```
Although both forms should work, I think this is the preferred form today. See https://docs.python.org/3/library/abc.html

network-config-analyzer/GenericTreeScanner.py
```diff
@@ -0,0 +1,95 @@
...
+        :param str path: the path of the file
+        :param stream: an IO-Text stream or Union of the file contents, depends on the scanner's type
+        """
+        decoded_stream = stream
+        if self.scanner_type == GenericTreeScanner.ScannerType.GitUrl:
+            decoded_stream = stream.decoded_content
+        load_type = None
+        if not self.np_scanner:
+            load_type = "safe"
+        yaml = YAML(typ=load_type)
```
zivnevo, line 65:
```suggestion
        yaml = YAML(typ='safe')
```
We should fix our tests, instead of giving-up safe loading. The fix is rather simple: adding quotes around selectors with '!'.
That is, change
```
      selector: !has(app)
```
into
```
      selector: "!has(app)"
```

network-config-analyzer/PeerContainer.py
```diff
@@ -111,37 +110,23 @@ def _set_namespace_list(self, ns_resources_list):
...
-                        for ns_code in code:
-                            if isinstance(ns_code, dict) and ns_code.get('kind') in {'NamespaceList', 'List'}:
-                                self.set_namespaces(ns_code)
-
             # load from live cluster
-            elif ns_resources == 'k8s':
+            if ns_resources == 'k8s':
                 self.load_ns_from_live_cluster()
+            else:
+                yaml_files = TreeScannerFactory.get_scanner(ns_resources).get_yamls()
```
zivnevo, line 125: Add protection against `get_scanner()` returning `None`

network-config-analyzer/PeerContainer.py
```diff
@@ -159,68 +144,22 @@ def _set_peer_list(self, peer_resources_list, config_name):
...
             # load from live cluster
-            elif peer_resources == 'calico':
+            if peer_resources == 'calico':
                 for peer_type in ['wep', 'hep', 'networkset', 'globalnetworkset']:
                     peer_code = yaml.load(CmdlineRunner.get_calico_resources(peer_type), Loader=yaml.SafeLoader)
                     self.add_eps_from_list(peer_code)
             elif peer_resources == 'k8s':
                 self.load_peer_from_k8s_live_cluster()
+            else:
+                yaml_files = TreeScannerFactory.get_scanner(peer_resources).get_yamls()
```
zivnevo, line 155: Add protection against `get_scanner()` returning `None`

network-config-analyzer/NetworkConfig.py
```diff
@@ -186,30 +183,14 @@ def _add_policies(self, doc, file_name, is_list=False):
...
-            for file in files:
-                if not file.endswith('.yaml') and not file.endswith('.yml') and not file.endswith('.json'):
-                    continue
-                file_with_path = os.path.join(root, file)
-                self.add_policies_from_file(file_with_path)
-
-    def add_policies_from_github(self, url):
-        yaml_files = GitScanner(url).get_yamls_in_repo()
+    def scan_entry_for_policies(self, entry):
+        yaml_files = TreeScannerFactory.get_scanner(entry, np_scanner=True).get_yamls()
```
zivnevo, line 187: Add protection against `get_scanner()` returning `None`

zivnevo (changes requested):
This turned out to be a real "duplicates killer". Well done.
A few small fixes and we are done.

zivnevo (approved):
ok to merge

PR #108, opened 2021-11-16
Tcp properties revisited

tanyaveksler:


network-config-analyzer/PortSet.py
```diff
@@ -113,178 +123,79 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
...
-                res_excluded_named_ports[port_name] = other.named_ports[port_name]
-        self.excluded_named_ports = res_excluded_named_ports
-
+        assert not self.excluded_named_ports
+        assert not isinstance(other, TcpProperties) or not other.named_ports
+        assert not isinstance(other, TcpProperties) or not other.excluded_named_ports
+        super().__isub__(other)
         return self
 
     def contained_in(self, other):
```
adisos, line 201: should we handle containment only without named ports?

network-config-analyzer/PortSet.py
```diff
@@ -99,7 +99,17 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
...
+
+        self.named_ports = {}  # a mapping from dst named port (String) to src ports interval set
+        self.excluded_named_ports = {}  # a mapping from dst named port (String) to src ports interval set
+        # assuming named ports are only in dest, not src
+        all_ports = PortSet.all_ports_interval.copy()
+        for port_name in dest_ports.named_ports:
+            self.named_ports[port_name] = source_ports.port_set
+        for port_name in dest_ports.excluded_named_ports:
+            # self.excluded_named_ports[port_name] = all_ports - source_ports.port_set
+            self.excluded_named_ports[port_name] = all_ports
```
adisos, line 112: should excluded named ports be mapped to source_ports as well? 
Can we simplify the named ports representation? (avoid the mapping to source ports?)

network-config-analyzer/PortSet.py
```diff
@@ -113,178 +123,79 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
...
-        res_excluded_named_ports = dict({})
-        for port_name in self.excluded_named_ports:
-            res_excluded_named_ports[port_name] = self.excluded_named_ports[port_name]
-        for port_name in other.named_ports:
-            if port_name in res_excluded_named_ports:
-                res_excluded_named_ports[port_name] -= other.named_ports[port_name]
-        self.excluded_named_ports = res_excluded_named_ports
-
+        assert not isinstance(other, TcpProperties) or not other.excluded_named_ports
+        super().__ior__(other)
```
adisos, line 181: It seems that the 'or' operation between objects with named ports is only used in k8s policy parsing of ports. (and it has no negative ports, and no source ports). should we simplify? add some documentation/assumptions?
(in the calico parsing is not using operators between objects of this type, only creates them).

adisos (commented):
hi @tanyaveksler ,
I've added some comments, with a few questions/thoughts whether we can further simplify the handling of the named ports, and whether we should correct a few aspects of the current version.

network-config-analyzer/PortSet.py
```diff
@@ -113,178 +123,79 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
...
-        for port_name in dest_ports.named_ports:
-            self.named_ports[port_name] = source_ports.port_set
-        for port_name in dest_ports.excluded_named_ports:
-            # self.excluded_named_ports[port_name] = all_ports - source_ports.port_set
-            self.excluded_named_ports[port_name] = all_ports
+        self.add_cube(cube, active_dims)
 
     def __bool__(self):
-        return bool(self.cubes_set) or bool(self.named_ports)
+        return super().__bool__() or bool(self.named_ports)
```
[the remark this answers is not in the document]
tanyaveksler: We cannot assume it. In add_connections() function bool() is called when building connections (during parsing), when they may still include named ports.

network-config-analyzer/PortSet.py
tanyaveksler: We cannot avoid keeping mapping of named ports to source ports, since in case of dst ports consisting of named ports only (no real numbers), the cube is not added (since dim_val for dst_ports will be empty in the cube):
From TcpLikeProperties.__init__
            has_empty_dim_value = False
            for dim_val in cube:
                if not dim_val:
                    **has_empty_dim_value** = True
                    break
            if not **has_empty_dim_value**:
                self.add_cube(cube, active_dims)

tanyaveksler: Discussed this with Ziv and he agreed that the most correct (and ugly :) way is to keep named_ports and excluded_named_ports as they are today

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -48,43 +58,37 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
...
-        for port_name in dest_ports.excluded_named_ports:
-            # self.excluded_named_ports[port_name] = all_ports - source_ports.port_set
-            self.excluded_named_ports[port_name] = all_ports
+                self.add_cube(cube, active_dims)
 
     def __bool__(self):
-        return bool(self.cubes_set) or bool(self.named_ports)
+        return super().__bool__() or bool(self.named_ports)
+
+    def get_simplified_str(self):
```
adisos, line 76: delete this method

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -120,155 +124,66 @@ def get_properties_obj(self):
...
-        for port_name in self.excluded_named_ports:
-            res_excluded_named_ports[port_name] = self.excluded_named_ports[port_name]
-        for port_name in other.excluded_named_ports:
-            if port_name in res_excluded_named_ports:
-                res_excluded_named_ports[port_name] |= other.excluded_named_ports[port_name]
-            else:
-                res_excluded_named_ports[port_name] = other.excluded_named_ports[port_name]
-        self.excluded_named_ports = res_excluded_named_ports
-
+        assert not self.excluded_named_ports
```
adisos, line 161: can replace the three assertions here with the following two assertions:
```
assert not self.has_named_ports()
assert not isinstance(other, TcpLikeProperties) or not other.has_named_ports()
```

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -120,155 +124,66 @@ def get_properties_obj(self):
...
-
-        res_excluded_named_ports = dict({})
-        for port_name in self.excluded_named_ports:
-            res_excluded_named_ports[port_name] = self.excluded_named_ports[port_name]
-        for port_name in other.named_ports:
-            if port_name in res_excluded_named_ports:
-                res_excluded_named_ports[port_name] -= other.named_ports[port_name]
-        self.excluded_named_ports = res_excluded_named_ports
-
+        assert not isinstance(other, TcpLikeProperties) or not other.excluded_named_ports
```
adisos, line 168: can also add the assertion: (after doing the update in the constructor as wrote on prev comment)
```
assert not self.excluded_named_ports
```

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -120,155 +124,66 @@ def get_properties_obj(self):
...
-        for port_name in self.excluded_named_ports:
-            res_excluded_named_ports[port_name] = self.excluded_named_ports[port_name]
-        for port_name in other.named_ports:
-            if port_name in res_excluded_named_ports:
-                res_excluded_named_ports[port_name] |= other.named_ports[port_name]
-            else:
-                res_excluded_named_ports[port_name] = other.named_ports[port_name]
-        self.excluded_named_ports = res_excluded_named_ports
-
+        assert not self.excluded_named_ports
```
adisos, line 183: can replace the 3 assertions with the following:
```
assert not self.has_named_ports()
assert not isinstance(other, TcpLikeProperties) or not other.has_named_ports()
```

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -277,19 +192,11 @@ def contained_in(self, other):
...
-                return False
-            if not self.named_ports[port_name].contained_in(other.named_ports[port_name]):
-                return False
-        for port_name in other.excluded_named_ports:
-            if port_name not in self.excluded_named_ports:
-                return False
-            if not other.excluded_named_ports[port_name].contained_in(self.excluded_named_ports[port_name]):
-                return False
-        return True
+        assert not self.named_ports
```
adisos, line 195: can replace the 4 assertions with the following:
```
assert not self.has_named_ports() and not other.has_named_ports()
```

adisos (changes requested):
Hi @tanyaveksler , I've added a few comments. 
I also noticed that in the function get_named_ports we should probably fix its current version and also add the excluded named ports there. (this would also require to fix the number of expected warnings in test: tests/calico_testcases/example_policies/testcase15-ports/testcase15-scheme.yaml as well. )

network-config-analyzer/TcpLikeProperties.py
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -28,7 +28,17 @@ def __init__(self, source_ports=PortSet(), dest_ports=PortSet(), methods=None, p
         :param MinDFA paths: The dfa of http request paths
         :param MinDFA hosts: The dfa of http request hosts
         """
-        self.cubes_set = CanonicalHyperCubeSet(TcpLikeProperties.dimensions_list)
+        super().__init__(TcpLikeProperties.dimensions_list)
+
+        self.named_ports = {}  # a mapping from dst named port (String) to src ports interval set
+        self.excluded_named_ports = {}  # a mapping from dst named port (String) to src ports interval set
+        # assuming named ports are only in dest, not src
```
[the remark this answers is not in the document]
tanyaveksler: Done

network-config-analyzer/TcpLikeProperties.py
```diff
@@ -9,7 +9,7 @@
 from PortSet import PortSet
 
 
-class TcpLikeProperties:
+class TcpLikeProperties(CanonicalHyperCubeSet):
```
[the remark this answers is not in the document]
tanyaveksler: Done

adisos (approved):
ok to merge

Issue #104, opened 2021-11-15
Avoid per-query code duplication in SchemeRunner


zivnevo:
Currently, the same code for running a query is pretty much duplicated for each query.
Maybe a generator can replace parts of the for loops in each query

shireenf-ibm:
Hi @zivnevo 
while unifying the queries functions , I realized an issue with the InterferesQuery.

When calling it from the CLExecute.py , the base_np_config is passed to it as its 2nd parameter as followed : 
`        full_result = NetworkConfigQuery.InterferesQuery(exclusive_network_policy, base_np_config).exec()
`

while in SchemeRunner.py the base_config is passed as its first param as following (def _run_interferes) : 
`            full_result = InterferesQuery(base_config, self._get_config(config), output_configuration).exec()
`

In my code i will use a one method for both command-line and scheme-runner calls as well the same call will be used for other queries (permits, forbids, containment ..) and since the other queries send the base config as the 2nd parameter , I prefer that too.. as follows: 
`        full_result = getattr(NetworkConfigQuery, query_type)(config, base_config, self.output_configuration).exec()
`
To solve the conflicts that are caused I think of 2 options:
-> make changes in the InterferesQuery.exec() 
-> or make changes to the "scheme" tests that may fail if we send the base config as the 2nd parameter 

would like to know your opinion in this

zivnevo:
> 

I think we can flip the parameters in the call from `CLExecute.py` (and fix relevant tests). This means we'll be checking if the given network-config interferes with the base network-config. It actually makes more sense.
Just make sure to update the documentation (in code and in `README.md`).

PR #110, opened 2021-11-28
refactoring queries runs

shireenf-ibm:
In this PR , the _run_<query_name> methods in SchemeRunner and query methods in CLexcute.
Instead created a new class (currently in a separate file) named NetworkConfigQueryRunner to handle the execution of the appropriate query (main method there is run_query, called from both the schemeRunner and CLexcute)
but still this code is not "beautiful" , because after generating the query name and calling its "exec" function we need to "analyze" its result.
exec method returns a "QueryAnswer" object result, but the final returned values (res and query_output) needs to be computed differently for each query (all these "computations" are based on
combinations of the members of the QueryAnswer).
So I think that these computations on the queryAnswer result should be done per QueryClass in the NetworkConfigQuery file.

"Brain storming":
we may change the queryAnswer class to contain the final "res" and "query_output". 
or may be add a sub class of QueryAnswer and handle the computations in a new method per <query> class 
so we can call this method after the exec , or create a wrapper method in each class that calls the exec then post process its result and returns the final results.
maybe even to make changes in the exec methods' computations
I would like to hear your ideas


network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -990,36 +992,38 @@ class InterferesQuery(TwoNetworkConfigsQuery):
     """
 
     def exec(self):
+        base_config = self.config2
+        other_config = self.config1
```
zivnevo, line 996: I'm not sure these aliases really help readability.
Later in the code `name1` and `name2` are used, and it's hard to remember which refer to which

zivnevo (changes requested):
Excellent work - code look much nicer now.
I have some thoughts on how to simplify things even further. Please see my comments below.

1. `_lower_camel_to_snake_case()` in `SchemeRunner.py` can be removed.
2. In `CLExecute.py`, let's not make `interferes` such an exception. We can cancel the (undocumented) feature of specifying a NetworkPolicy name as `exclusive_network_policy`. It is never used anyway.  This also requires removing the command-line test `named_policy_interferes`.
3. If possible, avoid wherever possible referring to query names by comparing strings (e.g., code like `if self.query_name in {'permits', 'forbids', 'interferes', 'containment'}`). Instead, add properties to the Query object itself, and use the `getattr(NetworkConfigQuery, query_type)` trick. This way you can even avoid the two dicts in `NetworkConfigQueryRunner`

network-config-analyzer/CLExecute.py
```diff
@@ -22,73 +22,30 @@ def __init__(self, base_ns_list=None, base_pod_list=None, ns_list=None, pod_list
...
+        :param str query_name: the name of the arg.query
         :param str np1_list_location: First set of policies
-        :param str np2_list_location: Second set of policies
-        :return: 0 if the sets of policies are semantically equivalent. 1 otherwise
+        :param str np2_list_location: Second set of policies (or the base_policies_list)
+        :return: result of executing the query
         :rtype: int
         """
         network_config1 = NetworkConfig(np1_list_location, self.peer_container, [np1_list_location])
         network_config2 = NetworkConfig(np2_list_location, self.peer_container, [np2_list_location])
```
adisos, line 39: I think there is an issue here to assign for both network configs the same peer_container.
for the case of semantic diff query , it was originally as follows (on master):
```
network_config1 = NetworkConfig(np1_list_location, self.base_peer_container, [np1_list_location])
network_config2 = NetworkConfig(np2_list_location, self.peer_container, [np2_list_location])
```

so should refer to both ` self.base_peer_container` and ` self.peer_container`

adisos (changes requested):
@shireenf-ibm I noticed there are issues with the output tests, so started to take a look at them. It seems there is a gap for expected output with a query of semantic diff from command line. If you try the following command that I've added below, you can see the output on master and on this branch are different. I've added a comment in one of the changed files where I suspect the issue is originated from.

 `python .\nca.py --semantic_diff ..\tests\example_policies\tests-different-topologies\policy_b.yaml --ns_list ..\tests\different_topologies\ns_list_b.json --pod_list ..\tests\different_topologies\pods_list_b.json --base_np_list ..\tests\example_policies\tests-different-topologies\policy_a.yaml --base_ns_list ..\tests\different_topologies\ns_list_a.json --base_pod_list ..\tests\different_topologies\pods_list_a.json`

network-config-analyzer/nca.py
```diff
@@ -79,24 +79,24 @@ def run_args(args):
...
 
     if args.semantic_diff:
-        return cl_execute.semantic_diff(args.semantic_diff, base_np_list)
+        return cl_execute.execute_pair_configs_query('semanticDiff', base_np_list, args.semantic_diff)
 
-    return cl_execute.sanity(args.sanity or 'k8s')
+    return cl_execute.execute_single_config_query('sanity', args.sanity or 'k8s')
 
 
 def nca_main(argv=None):
```
shireenf-ibm, line 102: the base_np_list and args.<queries> were switched un tis file for some queries in order to keep the "source" behavior
since I send the network_configs as an array in the clexecute file and take the first element of the array as the base_config.
The interferes was changed as discussed in [https://github.com/IBM/network-config-analyzer/issues/104]( ), the Readme was updated accordingly

network-config-analyzer/CLExecute.py
shireenf-ibm: Hi @adisos 
I fixed the above failing output.
In the new commit all failing output tests , are failing because of whitespaces and empty lines.
except for the cmd_line tests that include connectivity map query, I added the "config name" to the output, in order to be consistent with the output of the scheme files' queries' output (didn't affect the results).
so if this is OK for you, would like to know how to override the expected outputs safely in order to commit them too

adisos: hi @shireenf-ibm , it looks ok now, you can override expected output by running the following: 
`tests/run_all_tests.py --type=output --action=override_expected_output`

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1044,6 +1133,14 @@ def exec(self, only_captured):
...
                            ' do not intersect the connections allowed by ' + self.name2)
 
+    def compute_query_output(self, query_answer, cmd_line_flag=False):
+        if query_answer.bool_result:
+            query_output = self.config2.name + ' does not forbid connections specified in ' + \
+                                        self.config1.name + ':' + query_answer.output_explanation
+        else:
+            query_output = self.config2.name + ' forbids connections specified in ' + self.config1.name
+        return QueryOutput(query_answer.bool_result, query_output)
+
```
zivnevo, line 1143: Let's have a separate `ForbidsQuery` class, which calls `IntersectsQuery`. The same way as `PermitsQuery` calls `ContainmentQuery`.

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -24,6 +24,12 @@ class QueryAnswer:
     numerical_result: int = 0
 
 
+@dataclass
+class QueryOutput:
+    res: int = 0
+    query_output: str = ''
```
zivnevo, line 30: Do we really need a new class? I find it confusing.
Maybe we can just add `query_output` field to `QueryAnswer`...
`res` seems to be always equal to `QueryAnswer.numerical_result`

zivnevo (changes requested):
Nice progress.
Let's see if we can make it even cleaner.

network-config-analyzer/nca.py
```diff
@@ -77,32 +110,35 @@ def run_args(args):
...
         return SchemeRunner(args.scheme, args.output_format, args.file_out).run_scheme()
 
-    ns_list = args.ns_list or 'k8s'
-    pod_list = args.pod_list or 'k8s'
-    base_ns_list = args.base_ns_list or args.ns_list
-    base_pod_list = args.base_pod_list or args.pod_list
     base_np_list = args.base_np_list or 'k8s'
-    output_format = args.output_format or 'txt'
-    cl_execute = CLExecute(base_ns_list, base_pod_list, ns_list, pod_list, output_format, args.file_out, args.pr_url)
+    base_peer_container = PeerContainer(args.base_ns_list or args.ns_list, args.base_pod_list or args.pod_list)
```
zivnevo, line 114: It might be that both `args.base_ns_list`  and `args.ns_list` are not defined. In this case, the default should be `k8s`.
Same for `args.base_pod_list` and `args.pod_list`.

tests/fw_rules_tests/policies/expected_output/semantic_diff_with_different_topologies-scheme_output.csv
```diff
@@ -1,3 +1,4 @@
+
```
zivnevo, line 1: Why do we get this redundant new line?

tests/output_config_tests/expected_output_files/basic_connectivity_txt__stdout_output.txt
```diff
@@ -1,7 +1,6 @@
 global: cluster has 29 unique endpoints, 8 namespaces
 global: cluster has 29 unique endpoints, 8 namespaces
-
-final fw rules for query: :
+final fw rules for query: , config: output_config_tests/../example_policies/testcase8/testcase8-networkpolicy1.yaml:
```
zivnevo, line 3: why this diff?

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -613,13 +665,26 @@ def exec(self):
                                        explanation)
         return QueryAnswer(True, self.name1 + ' and ' + self.name2 + ' are semantically equivalent.')
 
+    @staticmethod
+    def compute_query_output(query_answer, cmd_line_flag=False):
```
zivnevo, line 669:
```suggestion
    def compute_query_output(query_answer, _):
```
(`cmd_line_flag` is unused)

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -48,6 +60,20 @@ def __init__(self, config, output_config_obj=None):
...
+        if only_explanation:
+            return res, query_answer.output_explanation
+        query_output = query_answer.output_result
+        if add_explanation:
+            query_output += query_answer.output_explanation
+        return res, query_output
+
+    @staticmethod
+    def get_query_type():
+        return QueryType.SingleConfigQuery
```
zivnevo, line 75: Should we also have a default query type for `TwoNetworkConfigsQuery`?

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1045,6 +1193,27 @@ def exec(self, only_captured):
...
+    def get_query_type():
+        return QueryType.ComparisonToBaseConfigQuery
+
+    def exec(self):
+        if not self.config1:
+            return QueryAnswer(False,'There are no NetworkPolicies in the given forbids config. '
+                                     'No traffic is specified as forbidden.')
+        return IntersectsQuery(self.config1, self.config2).exec(True)
+
+    def compute_query_output(self, query_answer, cmd_line_flag=False):
```
zivnevo, line 1208:
```suggestion
    def compute_query_output(self, query_answer, _):
```

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -11,6 +11,14 @@
 from ConnectivityGraph import ConnectivityGraph
 from OutputConfiguration import OutputConfiguration
 from Peer import PeerSet, IpBlock
+from enum import Enum
```
zivnevo, line 14: place this import with all the other standard imports (i.e., after `import itertools`)

zivnevo (changes requested):
It's getting nicer :)
A few more comments...

tests/fw_rules_tests/policies/expected_output/semantic_diff_with_different_topologies-scheme_output.csv
shireenf-ibm: 1. In NetworkConfigQueryRunner.py a '\n' is always added after calling execution of a single query 
`output += query_output + '\n'`
2. SemanticDiff is one of the 2 queries that support output formats other than .txt 
3. In this specific test (fw_rules_tests\policies\semantic_diff_with_different_topologies-scheme.yaml) the first query (semantic_diff_identical) has a "no policy" as its first config, 
so for the **.txt** format it returns a query_output = no_policy and policy are semantically equivalent. (printed at first line in semantic_diff_with_different_topologies-scheme_output.txt ), but returns an empty query_output for other formats (.csv, .yaml , .md ..) - so we see an empty line in these files 

If needed, I may add an if condition in the NetworkConfigQueryRunner.py before adding the  "+ '\n'"
or adjust the exec method of semanticDiff  query to return something that could be printed in the other formats instead of empty output.

tests/output_config_tests/expected_output_files/basic_connectivity_txt__stdout_output.txt
shireenf-ibm: When removing the "def _run_connectivity_map(self, configs_array, output_configuration)" from schemeRunner.py, I moved the following line :
`output_configuration.configName = config`
to the exec method of ConnectivityMapQuery:
       ` self.output_config.configName = self.config.name`
 (In order to keep printing the config name for connectivity map queries from scheme files as was done before the refactoring)
 As a result, the config name is printed now for connectivity map queries from both scheme and cmdline sources.

network-config-analyzer/NetworkConfigQuery.py
shireenf-ibm: True, Fixed 
I chose to put the "ComparisonToBaseConfigQuery" as its default.
the method was removed from the relevant classes

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -13,6 +14,13 @@
 from Peer import PeerSet, IpBlock
 
 
+class QueryType(Enum):
+    SingleConfigQuery = 0
+    ComparisonToBaseConfigQuery = 1
+    PairComparisonQuery = 2
+    WisePairComparisonQuery = 3
```
zivnevo, line 21:
```suggestion
    PairwiseComparisonQuery = 3
```

zivnevo (changes requested):
Just a small typo to fix and we are done

zivnevo (approved):
ok to merge

Issue #93, opened 2021-10-20
tests improvement 


adisos:
- [x] add some performance tests, to compare expected runtime per test, and alert for new run with bad/unexpected performance.
- [x] add some options to run tests per categories (e.g. istio / k8s / calico)
- [ ] Add a list of all tests that failed

zivnevo:
Shireen, this is also something to take a look at.
We can start with just measuring and reporting the time per test and per category.

shireenf-ibm:
>     * [ ]  add some options to run tests per categories (e.g. istio / k8s / calico)

Hi, 
is adding a new --flag to choose the category of tests you want to run will be helpful?

adisos:
> > ```
> > * [ ]  add some options to run tests per categories (e.g. istio / k8s / calico)
> > ```
> 
> Hi, is adding a new --flag to choose the category of tests you want to run will be helpful?

yes, a new flag sounds good.
It may be supported for tests of type 'general' or 'output'.
It's easier to start with 'general'. For the 'output' tests we may need to move the test files into relevant categories dirs.

shireenf-ibm:
>     * [ ]  Add a list of all tests that failed

How to output the list of failing file? into a new output file or to the existing log (stdout) under the summary ?

adisos:
> > ```
> > * [ ]  Add a list of all tests that failed
> > ```
> 
> How to output the list of failing file? into a new output file or to the existing log (stdout) under the summary ?

To the existing log under the summary

PR #114, opened 2021-11-30
part1 - comparing actual_run_time vs expected_run_time for a regressi…

shireenf-ibm:
…on run..
in case a test takes more than twice its expected run time, a warning will be printed in the "test's section" in the output_log

issue #93


adisos (changes requested):
hi @shireenf-ibm , looks good, thanks. I have a few ideas for improving this functionality:

1. Can add a new flag so that the script compares actual runtime to expected runtime only when requested by the flag.
2. Instead of printing warnings, create some output csv file with the list of tests with unexpected runtime, and for each such test show the actual runtime and expected runtime.
3. This runtime comparison should only be relevant for tests of type 'general'.

tests/run_all_tests.py
```diff
@@ -516,17 +561,23 @@ def main(argv=None):
...
     test_type = args.type
     action = args.action
     clean_out_files = not args.dont_clean_output_files
+    check_run_time = args.check_run_time
     if action != 'run_tests' and test_type != 'output':
         print(f'action: {action} is not supported with test type: {test_type}')
         sys.exit(1)
+    # if check_run_time and test_type != 'general':
+    #     print(f'check_run_time flag is not supported with test type: {test_type}')
+    #     sys.exit(1)
```
shireenf-ibm, line 577: Hi,
I kept this commented because I am not sure if it is necessary since I've added the check only for general tests.
would like to know your opinion

adisos: I think we can keep it because if someone uses this flag for unsupported test type, this would issue an error

adisos (changes requested):
hi @shireenf-ibm , looks good, thanks.
Just change the check that you have added to be un-commented as I wrote you in the comment.

adisos (approved):
ok to merge

PR #122, opened 2021-12-19
fix connection set issues

adisos:


network-config-analyzer/ConnectionSet.py
```diff
@@ -409,6 +416,19 @@ def add_all_connections(self, excluded_protocols=None):
             else:
                 self.allowed_protocols[protocol] = True
 
+    def is_all_connections(self):
```
zivnevo, line 419: I suggest a function `check_if_all_connections()` that will also take care of setting the `allow_all` flag and the `allowed_protocol` map. This will prevent duplicating code.

network-config-analyzer/ConnectionSet.py
```diff
@@ -409,6 +416,19 @@ def add_all_connections(self, excluded_protocols=None):
...
+        """
+        check if self allows all connections, and can be replaced with allow_all flag
+        :rtype: bool
+        """
+        num_protocols = ConnectionSet._max_protocol_num - ConnectionSet._min_protocol_num + 1
```
zivnevo, line 424: Don't we want to check if `allow_all` flag is already `True`?

tests/output_config_tests/expected_output_files/basic_semantic_diff_yaml__stdout_output.txt
```diff
@@ -12,15 +12,267 @@ global: cluster has 29 unique endpoints, 8 namespaces
...
+    - Protocol: 7
+    - Protocol: 8
+    - Protocol: 9
+    - Protocol: 10
+    - Protocol: 11
+    - Protocol: 12
+    - Protocol: 13
+    - Protocol: 14
+    - Protocol: 15
+    - Protocol: 16
```
zivnevo, line 33: We definitely need protocol ranges.
Also, checking if complement is shorter?

zivnevo (changes requested):
See comments

network-config-analyzer/ConnectionSet.py
adisos: done

tests/output_config_tests/expected_output_files/basic_semantic_diff_yaml__stdout_output.txt
adisos: done

zivnevo (approved):
looks good

PR #124, opened 2021-12-21
Issue 93 run tests per category

shireenf-ibm:
In this commit I've added the --category flag which indicates what set of tests to run.
category values can be: '' (default, all tests), k8s, istio, and calico.
In order to add the option to run tests by category, I realized that scheme files are organized this way:
1. **For tests type = general:** (which runs all scheme files existing under tests/*)
	- tests/calico_testcases includes scheme files under the category calico
	- tests/example_policies includes scheme files under the category k8s
	- tests/fw_rules_tests includes scheme files for all categories, when the name of the file determines the category
	- tests/istio_testcases includes scheme files under the category istio
	- tests/networkpolicy_checks_bad_path includes scheme files under the category k8s
	- tests/output_config_tests includes scheme files under the category k8s
2. **For tests type = output:** this type of tests runs only on scheme files of 2 sub-directories:
	tests/fw_rules_tests and tests/output_config_tests as specified in all_tests_spec.yaml
	when in these directories,
	 if the name of the scheme file starts with calico_, then it belongs to the calico category
	and when it starts with istio_ it belongs to the istio category
	otherwise it belongs to k8s.

3. **CmdLine tests** that exists, on file runs with tests type general, and another one for tests type output, both belong to k8s category.

So my idea was, when we define a category (using the --category flag): 
for each test file to determine if the file matches the category based on its location/name, and if there is a match the the test runs

but in order to get a general, clean code and without a "coded by hand" dirs names, I found it helpful to unite tests/example_policies, tests/example_podlist and tests/networkpolicy_checks_bad_path
under a one directory (k8s_testcases).
I also chose to rename the cmdline tests files, to start with k8s_ (so in code I can use the category variable, instead of 'k8s' - in case in the future istio/calico cmdline tests will be added )
I hope this refactoring is OK for you (if not, I can re-code it using conditions on the current directories)


tests/run_all_tests.py
```diff
@@ -486,12 +486,36 @@ def create_and_run_test_obj(self, test_queries_obj_list, expected_res):
...
+            return True
+        return False
+
+    def _test_file_matches_category_general_tests(self, test_file):
+        if self.category == '':
+            return True
+        if self.category + '_testcases' in test_file:
+            return True
+        if '_testcases' not in test_file:
+            return self._test_file_matches_category_output_tests(test_file)
```
adisos, line 506: why matching for output tests within the function for general tests on this case?

shireenf-ibm: general test run on all the subdirs of tests/ ,
which include <category>_testcases directories, each belong to 1 category. And the remaining directories ( tests/fw_rules_tests and tests/output_config_tests) which are also the directories that "output" test runs on, and contain scheme_files from all categories

adisos (changes requested):
hi @shireenf-ibm , looks very good, and the refactoring also makes sense. Just added a few small comments. Thanks!

adisos (approved):
ok to merge

shireenf-ibm:
after merging the "reafactoring k8s directory" into master I need to update some "github links" in the project to be consist with the new links.. fixing it

Issue #118, opened 2021-12-06
Fix security warnings related to REST API and DirScanner


zivnevo:
See the warnings here:
https://github.com/IBM/network-config-analyzer/security/code-scanning

Basically the warnings refer to running NCA as a REST server (with the `--daemon` flag).
In this mode, the user uploads the NetworkPolicies to analyze using a PUT request. The security problem is that the user can put a filename instead and somehow access the server's secured files.

### How to fix
In `RESTServer.py:95` avoid constructing a `NetworkConfig` instance using the `buffer` hack. Instead, use a `from_buffer` flag to the constructor.

### How to test
Follow the instructions [here](https://github.com/IBM/network-config-analyzer/blob/master/tests/test_rest_api.md)

PR #125, opened 2021-12-22
avoiding the buffer hack and using a flag instead

shireenf-ibm:
#118

### DIFF
--- a/network-config-analyzer/NetworkConfig.py
+++ b/network-config-analyzer/NetworkConfig.py
@@ -35,7 +35,7 @@ class NetworkConfig:
         Calico = 2
         Istio = 3
 
-    def __init__(self, name, peer_container, entry_list=None, config_type=None):
+    def __init__(self, name, peer_container, entry_list=None, config_type=None, from_buffer=False):
         """
         :param str name: A name for this config
         :param PeerContainer peer_container: The set of endpoints and their namespaces
@@ -51,6 +51,7 @@ class NetworkConfig:
         self.referenced_ip_blocks = None
         self.type = config_type or NetworkConfig.ConfigType.Unknown
         self.allowed_labels = set()
+        self.from_buffer = from_buffer
         peer_container.clear_pods_extra_labels()
         for entry in entry_list or []:
             self.add_policies_from_entry(entry)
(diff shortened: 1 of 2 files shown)


shireenf-ibm:
I followed the solution in #118, 
However by following the instructions [here](https://github.com/IBM/network-config-analyzer/blob/master/tests/test_rest_api.md) I got the expected results on both master branch and the current one.

zivnevo:
Try to directly use the flag in `NetworkConfig.__init__()` rather than putting it into a data member and using it later.

network-config-analyzer/NetworkConfig.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, name, peer_container, entry_list=None, config_type=None):
         self.referenced_ip_blocks = None
         self.type = config_type or NetworkConfig.ConfigType.Unknown
         self.allowed_labels = set()
+        if from_buffer:
+            entry_list = 'buffer: ' + entry_list
         peer_container.clear_pods_extra_labels()
         for entry in entry_list or []:
             self.add_policies_from_entry(entry)
```
zivnevo, line 58:
```suggestion
        peer_container.clear_pods_extra_labels()
        if from_buffer:
            entry_list = 'buffer: ' + entry_list
        else:
            for entry in entry_list or []:
                self.add_policies_from_entry(entry)
```
Sorry, I wasn't clear enough. Here is my suggestion.

zivnevo (changes requested):
Sorry, I wasn't clear enough. Here is my suggestion.

network-config-analyzer/NetworkConfig.py
zivnevo, line 58:
```suggestion
        peer_container.clear_pods_extra_labels()
        if from_buffer:
            self._add_policies(entry, 'buffer', True)
        else:
            for entry in entry_list or []:
                self.add_policies_from_entry(entry)
```

zivnevo (changes requested):
Corrected suggestion

zivnevo:
ok, so this fails too. How about using two different parameters to `NetworkConfig.__init__()`?
That is, something like:
```
    def __init__(self, name, peer_container, entry_list=None, config_type=None, buffer=None):
```
And then, call the ctor from `RestServer.py` with `entry_list` set to `None` and `buffer` set to `entry`.

shireenf-ibm:
> ok, so this fails too. How about using two different parameters to `NetworkConfig.__init__()`? That is, something like:
> 
> ```
>     def __init__(self, name, peer_container, entry_list=None, config_type=None, buffer=None):
> ```
> 
> And then, call the ctor from `RestServer.py` with `entry_list` set to `None` and `buffer` set to `entry`.

working :)

zivnevo (approved):
hooray

PR #138, opened 2022-01-17
expectedOutput support with general tests type

shireenf-ibm:
* In this PR , expectedOutput option is supported for both scheme files and cmd-line
* It runs with general tests type.
* I still have few things to do (mainly in the "run_all_tests.py -  kept the OutputTest class commented in the meanwhile).

By adding the expectedOutput file name to the query's attributions :
 
-> comparing the results is done between the given file content and the actual "query output" computed by the QueryRunner
(there will be no comparison with the stdout contents - message from other places (e.g warning from the np parser) will not be compared and shouldn't be in the query's expected output file) 
-> when adding the expectedOutput to a query, if its format isn't txt, the user should include OutputFormat parameter.
overriding the format from outside will not be relevant any more in this case.
We had this option in the old "Output" section of all_tests_spec.yaml, it was relevant for the  tests/output_config_tests/override_out_format_only/test_output_config_connectivity_queries_override_out_format-scheme.yaml test.
**Question:** Since it is not relevant now , should I remove this test or implement a new flag in run_all_tests (--override_test_format), however the user will be responsible to use the correct expectedOutput file format in this case?!
-> **Question:** Since now the expectedOutput is per query , should we keep the actual output files per scheme, or change them to be per query too?


adisos (changes requested):
Hi @shireenf-ibm , looks very good, thanks!
A few things that I think we should consider:

- [x] update file .github/workflows/test-push.yml : delete the run of tests/run_all_tests.py with output flag 
- [x] should limit the support of expectedOutput only to some query type (connectivityMap and semanticDiff) [because for other queries it may be non deterministic]
- [x] can implement some flag for run_all_tests that would allow to auto create/update the expected output file of some new tests or existing tests? (instead of reporting a failure)?
(auto create: add missing file, auto update: override expected output file)
- [x] what if someone write for the query expectedOutput without expected: 0 ? (currently failed test is not reported as failure on this case) [maybe should have by default expected=0 for any test with expectedOutput arg?]
- [ ] actual output files: could be redundant now, because we do not use output redirection to files when tests are run, instead the output per query goes to stdout and can be compared with expected output file from query...
- [x] migration of fw_rules tests:  fw_rules_csv_md : only poc1 is containing the formats csv,md,dot (missing the smeantic_diff tests with those formats)
fw_rules_txt_and_yaml: to preserve what has been tested so far, should have txt and yaml for each test under fw_rules 
- [x] output config tests: maybe we should consider a new test type for them, and not run them as general tests? Their goal was to test combinations of outputConfig args (output path, output format) + combinations of those cli args that can override them.
- for outputConfig outPath : make sure the expected file is created with expected content 
- for outputConfig outFormat: make sure that specifying it does affect the output format (at least one check for each possible type)
- for cli arg outPath: make sure the expected file is created with expected content (regardless of outputConfig outPath if exists..)
- for cli arg outFormat: make sure the output format is correct (regardless of outputConfig outPath)
- check combinations of overriding only one arg (each one separately) or both

shireenf-ibm:
>     * [x]  what if someone write for the query expectedOutput without expected: 0 ? (currently failed test is not reported as failure on this case) [maybe should have by default expected=0 for any test with expectedOutput arg?]

Instead of defaulting expected to 0 , returning a "comparing_err" flag from the queryRunner, and adding it to the global_res in schemeRunner/nca (cmdline) ... (because there are tests without expected but their actual res is > 0, which will cause a test failure when defaulting expected to 0)
 
>     * [x]  actual output files: could be redundant now, because we do not use output redirection to files when tests are run, instead the output per query goes to stdout and can be compared with expected output file from query...

I kept actual output files only for the output files that are given in OutputConfig's args / cmdline arg / overrided by the spec file.
deleted all "stdout" redirected actual output files

>     * [x]  output config tests: maybe we should consider a new test type for them, and not run them as general tests? Their goal was to test combinations of outputConfig args (output path, output format) + combinations of those cli args that can override them.

- In the committed files , I added a new class OutputConfigTests (--type=output_configuration)
But this class behaves exactly as the GeneralTest behaves.
the only difference is in the spec file (all_tests_spec.yaml) - when adding the "override" args.
these args (format/out file) are passed to nca from run_all_tests.py. (nca/schemeRunner are responsible to handle them- didn't have to touch them).
- Since the OutputConfigTests behave like the General, I wrote another version without the new class (attached).

So we may choose the "committed"/ the "attached" version

 - In both solutions I had to add a new "parameter" to the spec file called "excluded_dirs" >> used under the general scheme tests.
The reason for adding the new parameter to exclude the output_config_tests directory from running all scheme files is that the expected output files is now in the "scheme" files (per query) and for tests that we'll override their format , the expected is already written with the overriden format, so we can't run them without the "out_format_arg"  
 
> 
>     * for outputConfig outPath : make sure the expected file is created with expected content
> 
>     * for outputConfig outFormat: make sure that specifying it does affect the output format (at least one check for each possible type)
> 
>     * for cli arg outPath: make sure the expected file is created with expected content (regardless of outputConfig outPath if exists..)
> 
>     * for cli arg outFormat: make sure the output format is correct (regardless of outputConfig outPath)
>
>     * check combinations of overriding only one arg (each one separately) or both

Since the nca/schemeRunner handle the override format/file args -> all outputs are correct for sure.
since the nca and schemeRunner take these parameters and override them in the query's OutputConfig ... 
and in the queryRunner we write the out file and compare the output with same "query_output" variable
[all_tests_spec_without_output_config_test.yaml.txt](https://github.com/IBM/network-config-analyzer/files/7941058/all_tests_spec_without_output_config_test.yaml.txt)
[run_all_tests_without_output_config_class.py.txt](https://github.com/IBM/network-config-analyzer/files/7941065/run_all_tests_without_output_config_class.py.txt)

adisos (changes requested):
hi @shireenf-ibm , a few more comments:

- [ ] update documentation of scheme files
- [ ] update workflow to override tests output
- [ ] remove the compare/update arguments from nca_main, scheme_runner, NetworkConfigQueryRunner (add some new singleton object instead, so it can get updated from the run_all_tests directly).
- [ ] in scheme runner, if there is a comparing_err, add warning as well
- [ ] output_config tests can run as general tests

network-config-analyzer/SchemeRunner.py
```diff
@@ -35,7 +37,7 @@ def __init__(self, scheme_file_name, output_format=None, output_path=None):
             if not isinstance(self.scheme, dict):
                 self.syntax_error("The scheme's top-level object must be a map")
 
-    def _get_input_file(self, given_path):
+    def _get_input_file(self, given_path, out_flag=False):
```
adisos, line 40: update docstring

shireenf-ibm:
> hi @shireenf-ibm , a few more comments:
> 
>     * [x]  update documentation of scheme files
> 
>     * [x]  update workflow to override tests output
> 
>     * [x]  remove the compare/update arguments from nca_main, scheme_runner, NetworkConfigQueryRunner (add some new singleton object instead, so it can get updated from the run_all_tests directly).
> 
>     * [x]  in scheme runner, if there is a comparing_err, add warning as well
> 
>     * [x]  output_config tests can run as general tests

in addition to the points above, the following new change added:
Since the "actual_output_files" we have now are the files written by "OutputConfiguration"  >>
I moved the handling of these output files from run_all_tests.py to the OutputConfiguration.py (using the new singleton class)

adisos (changes requested):
a few more comments

adisos (commented):
hi @shireenf-ibm , I'm still confused about the output config tests. If we do not compare output files that are created from out_file arg (and only compare output per query), why do we still need to preserve actual output files, or even run with those out_file flags?
Do we have any unique functionality being tested in output_config, or is it redundant? 
Should we move the cli tests with expected output outside the output_config tests?

shireenf-ibm:
hi @adi,

> Do we have any unique functionality being tested in output_config, or is it redundant?

  we don't have any unique functionality when overriding the "output_cofig_tests" args.
The users have the option to choose the format/out_file in both scheme (in the outputConfiguration) and cmdline tests (-o -f)
so in my opinion there is no special additive value by overriding the out file/format in the tests_spec file.

> Should we move the cli tests with expected output outside the output_config tests?

yes, we might even merge the cli tests under output_config tests with the k8s_cmdline_tests.yaml , 
both files include same "base "tests, the output_config adds variations of new formats /out files

> why do we still need to preserve actual output files, or even run with those out_file flags?
even though we don't use the actual output files in our code anymore.
I think it is good to keep the (scheme outputConfiguration: outfile)/(cli -f) as an option for the user, to choose if he wants to keep his tests' output. (if he want to use them "outside" the nca run) 
for example if he want to write a new test on the same networkpolicies and want to analyze what happens in each case or if he wants to compare nca's output with other sources ...

adisos:
> hi @adi,
> 
> > Do we have any unique functionality being tested in output_config, or is it redundant?
> 
> we don't have any unique functionality when overriding the "output_cofig_tests" args. The users have the option to choose the format/out_file in both scheme (in the outputConfiguration) and cmdline tests (-o -f) so in my opinion there is no special additive value by overriding the out file/format in the tests_spec file.
> 
> > Should we move the cli tests with expected output outside the output_config tests?
> 
> yes, we might even merge the cli tests under output_config tests with the k8s_cmdline_tests.yaml , both files include same "base "tests, the output_config adds variations of new formats /out files
> 
> > why do we still need to preserve actual output files, or even run with those out_file flags?
> 
> even though we don't use the actual output files in our code anymore. I think it is good to keep the (scheme outputConfiguration: outfile)/(cli -f) as an option for the user, to choose if he wants to keep his tests' output. (if he want to use them "outside" the nca run) for example if he want to write a new test on the same networkpolicies and want to analyze what happens in each case or if he wants to compare nca's output with other sources ...

Thanks. In my last question I meant to ask whether we need to keep the tests that we wrote with this out-file flag, and why we need to preserve the "actual output files" from those tests, if we do not compare them to anything.
To simplify, if we do not find much added value for the output config tests, we can remove those tests, and merge the cli tests as you suggested. (and then can also simplify the run_all_tests).
The output config tests were originally meant to test both output comparison and behavior with outputConfig flags from scheme and/or cli args and their combinations. (their scheme files are basically duplication of regular tests scheme files, only with variations of the output config args). We may add some unit tests instead to make sure that the overriding of those args work as expected, and that outputConfig writes output to files as expected, but that may be for another issue.

adisos (changes requested):
hi @shireenf-ibm , we should keep the support for overriding output configuration args (in schemeRunner and nca.py).
We are only removing the output_config tests.

shireenf-ibm:
committed the remove of output_configuration_tests and its "support" in code.

for now: still keeping the "actual_output_files" , I don't see much value by preserving them either, but if it can be used as an indicator/example for any user so might should be available, especially that are written only when the dont_clean_output_files flag used.

shireenf-ibm:
> hi @shireenf-ibm , we should keep the support for overriding output configuration args (in schemeRunner and nca.py). We are only removing the output_config tests.

ok then re-writting the files

shireenf-ibm:
> > hi @shireenf-ibm , we should keep the support for overriding output configuration args (in schemeRunner and nca.py). We are only removing the output_config tests.

I am confused a bit.
- in cmdline tests the format and output args are still supported.
- in scheme runner, the only option to override the format/output args was by adding these args to the tests_spec file.
(now the only option is to add them in the the output configuration section of each query in the scheme file).

so by "resetting the last changes from my last committs", we'll keep the out_file and out_format supporting from test spec (means also the changes in run_all_tests.py in last commit will be  reset too). am I right?

network-config-analyzer/nca.py
```diff
@@ -69,20 +69,23 @@ def _g_tick():
...
     :param PeerContainer peer_container: set of peers
     :param OutputConfiguration output_config: dict object
+    :param str expected_output: a file path to the expected output
     :rtype: int
     """
     network_config1 = NetworkConfig(np1_list, peer_container, [np1_list])
-    return NetworkConfigQueryRunner(query_name, [network_config1], output_config).run_query() > 0
+    res, comparing_err = NetworkConfigQueryRunner(query_name, [network_config1], expected_output,
+                                                  output_config).run_query()
+    return res + comparing_err
```
adisos, line 85: Do you think we need the older version of  `(res > 0) + comparing_err` here because we call `run_query `with `cmd_line_flag=False` on a single config query, whereas for pair configs query we call it with `cmd_line_flag=True`?
I noticed that ` compute_query_output returns` the `bool_result` if `cmd_line_flag `is `True `for semantic diff query. So if the return value for cli queries is the boolean results, I assume we do need the `res>0` version for the single config query?

adisos (changes requested):
looks good, added a few more small comments + question.

network-config-analyzer/nca.py
shireenf-ibm: I guess when I wrote the res > 0 , I really  had a case of cmdline query that returned a numerical res = 2 but its expected was 1 , so I concluded to make it a boolean expr (res > 0).
but now we don't have such case, so I am not sure if we still need the res > 0.
( maybe then the test returned 2 instead of 1 because of the changes I was making to code and  I had a wrong conclusion )

adisos: Which test returned a numerical res = 2  but its expected res was 1? 
I tried to check a different cli query based on a scheme test from `tests\calico_testcases\example_policies\calico-demo\sanity3-scheme.yaml`: After converting this test to a cli query, I see that on this branch the return value is 2, whereas on master branch the return value is True.
I think we should be consistent with return values as before this new feature of "expected output".

network-config-analyzer/SchemeRunner.py
```diff
@@ -175,21 +177,27 @@ def run_queries(self, query_array):
...
+                                                                  output_config_obj, self.network_configs).run_query()
+
             if 'expected' in query:
                 expected = query['expected']
                 if res != expected:
                     self.warning(f'Unexpected result for query {query_name}: Expected {expected}, got {res}\n', query)
                     self.global_res += 1
+            if comparing_err != 0:
+                self.warning(f'Unexpected output comparing result for query {query_name} ')
+                self.global_res += comparing_err
```
adisos, line 203:
```suggestion
                self.global_res += 1
```
To be more consistent with counting unexpected results

network-config-analyzer/NetworkConfigQueryRunner.py
```diff
@@ -57,9 +59,15 @@ def run_query(self, cmd_line_flag=False):
...
+        comparing_err = 0
         self.output_configuration.print_query_output(query_output, formats)
-        return res
+        if self.expected_output_file is not None:
+            if self.query_name in {'ConnectivityMapQuery', 'SemanticDiffQuery'}:
+                comparing_err = self._compare_actual_vs_expected_output(query_output)
+            else:
+                print(f'Warning: expectedOutput is not relevant for {self.query_name}. '
+                      'Output compare will not occur')
+        return res, comparing_err
```
adisos, line 70: Update in docstring the rtype for `run_query ` to `(int, int)`

adisos (commented):
a few more comments

network-config-analyzer/nca.py
shireenf-ibm: I can't remember which test it was, but I agree that its better to keep res > 0 
(even though I hesitated yesterday), I believe then I was much concentrating and had much more thinking on it.
If it gave True on master , then the "old" CLExecute file dealt with queries results as bool, so we should keep it that way

adisos:
hi @shireenf-ibm , as we discussed, let's update the following:

- [ ] Return value for cli query to be a number between 0 to 3, represented by 2 bits: first bit for original return value, and second bit for the output comparison result.
Make sure that original return value was indeed boolean for all cli queries (including equiv / 2waycontainment).
- [ ] Update documentation for the return value of cli query and scheme file

shireenf-ibm:
> hi @shireenf-ibm , as we discussed, let's update the following:
> 
>     * [x]  Return value for cli query to be a number between 0 to 3, represented by 2 bits: first bit for original return value, and second bit for the output comparison result.
>       Make sure that original return value was indeed boolean for all cli queries (including equiv / 2waycontainment).
> 
>     * [x]  Update documentation for the return value of cli query and scheme file

I read how the result of each cmd-line query is computed and wrote them to an excel table.
all cli pair configs queries return bool result.
single query **sanity** may return a result > 1 , thats the reason of res > 0 .
[command_line_query_results.xlsx](https://github.com/IBM/network-config-analyzer/files/8023751/command_line_query_results.xlsx)

shireenf-ibm:
I found a new bug when running single cli test with expected output file  (connectivity and semantic diff) 
The bug is with printing the config name in query output explanations.
the nca cmd line runs with absolute path.
but the config.name in code is the relative path from the "tests" dir. 
I can fix this by printing the basename of the file instead of paths, in the relevant output sentences. -- fixing it

docs/SchemeFileFormat.md
```diff
@@ -78,3 +79,6 @@ For example: `my_set/prod_ns/deny_all_policy`
 * _vacuity_ - Count of vacuous sets
 * _sanity_ - Count of sanity issues
 * _allCaptured_ - Count of non-captured pods
+
+#### Exit code meaning :
+The exit code of running a scheme-file represents the number of queries that their result did not match the given expected result or/and their output did not match the given expected output file contents.
```
adisos, line 84: Also counting mismatches for expectedWarnings and expectedError.

adisos:
> > hi @shireenf-ibm , as we discussed, let's update the following:
> > ```
> > * [x]  Return value for cli query to be a number between 0 to 3, represented by 2 bits: first bit for original return value, and second bit for the output comparison result.
> >   Make sure that original return value was indeed boolean for all cli queries (including equiv / 2waycontainment).
> > 
> > * [x]  Update documentation for the return value of cli query and scheme file
> > ```
> 
> I read how the result of each cmd-line query is computed and wrote them to an excel table. all cli pair configs queries return bool result. single query **sanity** may return a result > 1 , thats the reason of res > 0 . [command_line_query_results.xlsx](https://github.com/IBM/network-config-analyzer/files/8023751/command_line_query_results.xlsx)

great, thanks for this check

adisos (changes requested):
Looks good, added a few small comment about documentation.
Do you think we should document somewhere the bool result for cli query (as you analyzed in your spreadsheet)? currently we only document what is the numerical result for query within scheme file.

docs/SchemeFileFormat.md
zivnevo: It will be nice to have a documentation of the return value from cli queries.
Let's add a new md file in `docs`, with the same content as the excel file.

adisos (approved):
ok to merge

Issue #133, opened 2022-01-04
connectivity map enhancement


adisos:
add an option to have the connectivity map output by deployments as endpoints instead of pods as endpoints.
(consider as outputConfig option, and another flag for cli options).

Issue #131, opened 2022-01-03
additional tests improvements


adisos:
Improving output tests: consider how we can extend scheme files to have per query a new attribute 'expectedOutput' with a path to a file containing expected output. (currently we only have 'expected' with an integer result).
It is probably mostly relevant for connectivity map and semantic diff queries.
consider also the following:
Can we migrate the current output tests to be using this new attribute?
(Currently we have some kind of duplication for scheme files and network topologies in order to test their output. e.g. tests/fw_rules_tests/policies/istio-bookinfo-test-request-attrs-scheme.yaml  and tests/istio_testcases/example_policies/bookinfo-demo/bookinfo-test-request-attrs-scheme.yaml )

PR #140, opened 2022-01-19
issue #133 : add connectivity map option connectivityByDeployments

adisos:
- I will add some tests after the merge from issue #131, to avoid conflicts.
- I have added the outputConfig option "connectivityByDeployments". Should we add some new cli option, or is it sufficient to control through scheme files?


network-config-analyzer/ConnectivityGraph.py
```diff
@@ -52,29 +65,24 @@ def get_connectivity_dot_format_str(self):
...
+            peer_name, is_ip_block = self._get_peer_name(peer)
+            peer_color = "red2" if is_ip_block else "blue"
+            peer_lines.append(f'\t\"{peer_name}\" [label=\"{peer_name}\" color=\"{peer_color}\" fontcolor=\"{peer_color}\"]\n')
 
-        edge_lines = []
+        edge_lines = set()
         for connections, peer_pairs in self.connections_to_peers.items():
             for src_peer, dst_peer in peer_pairs:
-                if src_peer != dst_peer and not connections.allow_all:
+                if src_peer != dst_peer and not connections.allow_all and connections:
```
adisos, line 75: I think we should avoid adding edges with no connections.
Do we want to keep not adding edges with 'allow all'?

zivnevo: Yes, we should avoid adding edges with no connections.
We do want to keep edges with 'allow all'. Why did we skip them anyway?

adisos: I'm not sure why they were skipped.. I will change it now.

network-config-analyzer/ConnectivityGraph.py
```diff
@@ -52,29 +65,24 @@ def get_connectivity_dot_format_str(self):
             output_result += f'\tHEADER [shape="box" label=< <B>{self.output_config.queryName}/{self.output_config.configName}</B> > fontsize=30 color=webmaroon fontcolor=webmaroon];\n'
         peer_lines = []
```
zivnevo, line 66: Don't we want a set here too?

zivnevo (changes requested):
ToDos:
- [ ] A cli flag to turn on connectivity-by-deployments
- [ ] Tests (waiting for updates to testing scheme)

network-config-analyzer/ConnectivityGraph.py
adisos: right -- fixed.

Issue #144, opened 2022-01-24
Parallelize test-push workflow


zivnevo:
Use jobs instead of steps + use caching to skip duplicate work

zivnevo:
Done

PR #145, opened 2022-01-24
Issue #144 - parallel docker testing

zivnevo:


zivnevo:
This is the first (and easier) phase - docker doesn't need the `setup-python` step as it is done inside the package

adisos (approved):

PR #146, opened 2022-01-24
Issue #144 - parallel testing steps + a reusable action to setup NCA env

zivnevo:
Also, fixing versions in requirements.txt for caching to work.


zivnevo:
Every test category now runs in a job of its own, so they can all run in parallel.
The caching provided by `setup-python` action works well. `pip install` runs in just a few seconds.
There is no longer a required test call `test`. I will fix that after merging.

adisos:
> @adisos - should we wait for the changes from PR #138 before merging?

yes, let's wait for the changes from this PR (the output tests are expected to change and run as general tests).

adisos (approved):

PR #152, opened 2022-01-25
Service resource handling - initial implementation

tanyaveksler:
This is my initial service implementation, I have a few TODOs and questions. Also, I haven't created an interesting service resource yaml for testing, that corresponds to pods list. Will do it on Sunday.


network-config-analyzer/K8sYamlParser.py
```diff
@@ -0,0 +1,336 @@
...
+from K8sNamespace import K8sNamespace
+from K8sService import K8sService
+from CmdlineRunner import CmdlineRunner
+from GenericTreeScanner import TreeScannerFactory
+
+
+class K8sYamlParser(GenericYamlParser):
+    """
+    A parser for k8s NetworkPolicy objects
+    """
```
zivnevo, line 20: Fix class description

network-config-analyzer/K8sService.py
```diff
@@ -0,0 +1,79 @@
...
+            return self.name == other.name
+        return NotImplemented
+
+    def __hash__(self):
+        return hash(self.name)
+
+    def __str__(self):
+        return self.name
+
+    def set_type(self, type):
```
zivnevo, line 53: `type` is a Python built-in function - better rename

zivnevo (changes requested):
Thanks Tanya. Overall looks good.
Some general comments:
- [x] For now, let's use the `pod_list` switch and scheme-key to scan for services too. Otherwise we'll end up with a switch for every k8s/Istio/calico resource. We do plan to introduce a new "catch-all" switch for all resources in the future (issue #153)
- [x] `K8sYamlParser` is in fact a general base class, providing common utilities for parsing K8s resources. It looks a bit weird that it has a service-specific code. I think we want a new class: `K8sServiceParser`.
- [x] It's not NCA's job to check the syntax of Service resources. Let's mostly assume their syntax is correct, and simplify the code accordingly (of course, the code should never crash if the syntax is bad. Use safe methods such as `dict.get()`).

network-config-analyzer/K8sYamlParser.py
tanyaveksler: Done

network-config-analyzer/K8sService.py
tanyaveksler: Done

network-config-analyzer/SchemeRunner.py
```diff
@@ -82,14 +83,22 @@ def _add_config(self, config_entry, peer_container_global):
 
         ns_list = config_entry.get('namespaceList')
         pod_list = config_entry.get('podList')
-        if ns_list or pod_list:  # a local resource file exist
+        srv_list = config_entry.get('serviceList')
```
adisos, line 86: We want to avoid adding this new entry of `serviceList`, and instead have the services within the current `podList`

tests/fw_rules_tests/policies/test1-scheme.yaml
```diff
@@ -1,5 +1,6 @@
 namespaceList: ../podlist/ns_list.json
 podList: ../podlist/pods_list.json
```
adisos, line 2: podList can point to a directory with files containing both pod list and service list.

adisos (changes requested):
hi @tanyaveksler , please refer again to Ziv's first comment:  "let's use the pod_list switch and scheme-key to scan for services too."

tests/fw_rules_tests/policies/test1-scheme.yaml
tanyaveksler: I am not sure I understand how you want to get services from the pod list. If podList points to a directory, there may be many resource files there. We need to specify specific pods and services files.

adisos: Let's create a new directory with only one pods list file and only one services file. If pods_list in the scheme file points to this directory, nca should parse all the relevant yaml resources from this dir.

network-config-analyzer/SchemeRunner.py
```diff
@@ -137,13 +142,16 @@ def run_scheme(self):
...
-        allowed_keys = {'networkConfigList': 1, 'namespaceList': 0, 'podList': 0, 'queries': 0}
+        allowed_keys = {'networkConfigList': 1, 'namespaceList': 0, 'podList': 0, 'serviceList': 0, 'queries': 0}
         self.check_fields_validity(self.scheme, 'scheme', allowed_keys)
 
         # global resource files
         pod_list = self._handle_resources_list(self.scheme.get('podList', 'k8s'))
         ns_list = self._handle_resources_list(self.scheme.get('namespaceList', 'k8s'))
         peer_container = PeerContainer(ns_list, pod_list)
+        # look for service resources under 'pod_list' files
+        services = K8sServiceYamlParser.parse_service_resources(pod_list, peer_container)
```
adisos, line 153: Why 2 calls for `parse_service_resources` from scheme runner?

adisos (changes requested):
Hi @tanyaveksler , added a few comments to start with.

network-config-analyzer/K8sService.py
```diff
@@ -0,0 +1,79 @@
...
+            # a target port may be given either as a number or as a string (named port)
+            if isinstance(target_port, int):
+                self.target_port_number = target_port
+            else:
+                self.target_port_name = target_port
+            self.protocol = protocol
+            self.name = name
+
+    def __init__(self, name):
+        self.name = name
```
adisos, line 35: Where it the namespace of the service?

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            service.set_type(K8sService.ServiceType.ClusterIP) # the default type
+
+        selector = service_spec.get('selector')
+        if selector:
+            for key, val in selector.items():
+                self.check_label_key_syntax(key, selector)
+                self.check_label_value_syntax(val, key, selector)
+                service.add_selector(key, val)
+                service.target_pods |= peer_container.get_peers_with_label(key, [val])
```
adisos, line 66: `get_peers_with_label` should be restricted to the namespace of the service

adisos (commented):
a few more comments

tests/fw_rules_tests/policies/test1-scheme.yaml
```diff
@@ -1,5 +1,6 @@
...
+serviceList: ../podlist/kubernetes-manifests.yaml
```
adisos, line 3:
```suggestion
```

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,157 @@
...
+        api_version = srv_object.get('apiVersion')
+        if api_version != 'v1':
+            return None  # apiVersion is not properly set
+        metadata = srv_object.get('metadata')
+        if not metadata:
+            return None
+        srv_name = metadata.get('name')
+        if not srv_name:
+            return None
+        self.check_dns_subdomain_name(srv_name, metadata)
```
adisos, line 44:
```suggestion
```
Can remove this check of valid resource name, as these checks are only applied to policy resources (similarly, pod resource name is not checked).

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,157 @@
...
+                if not port_id:
+                    continue
+                self.check_port_syntax(port_id, False, 'Service port')
+                target_port = port.get('targetPort')
+                if not target_port:
+                    target_port = port_id
+                else:
+                    self.check_port_syntax(target_port, True, 'Service targetPort')
+                name = port.get('name', '')
+                if name:
```
adisos, line 81:
```suggestion
```

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,157 @@
...
+                    self.check_dns_label_name(name, srv_object)
```
adisos, line 82:
```suggestion
```
Can remove this check, as nca should not check the syntax of service resources

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,157 @@
...
+                service.add_selector(key, val)
+                service.target_pods |= peer_container.get_peers_with_label(key, [val])
+
+        ports = service_spec.get('ports')
+        if ports is not None:
+            for port in ports:
+                port_id = port.get('port')
+                if not port_id:
+                    continue
+                self.check_port_syntax(port_id, False, 'Service port')
```
adisos, line 74: can we remove port syntax check to simplify, and assume it's correct?

adisos (commented):
added a few more comments.

network-config-analyzer/SchemeRunner.py
```diff
@@ -137,13 +142,16 @@ def run_scheme(self):
         :return: The number of queries with unexpected result + number of configs with unexpected number of warnings
         :rtype: int
         """
-        allowed_keys = {'networkConfigList': 1, 'namespaceList': 0, 'podList': 0, 'queries': 0}
+        allowed_keys = {'networkConfigList': 1, 'namespaceList': 0, 'podList': 0, 'serviceList': 0, 'queries': 0}
```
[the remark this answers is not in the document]
tanyaveksler: Done

network-config-analyzer/K8sService.py
```diff
@@ -0,0 +1,79 @@
...
+        self.labels = {}  # Storing the namespace labels in a dict as key-value pairs
```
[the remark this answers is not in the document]
tanyaveksler: Removed labels

network-config-analyzer/SchemeRunner.py
```diff
@@ -87,9 +88,13 @@ def _add_config(self, config_entry, peer_container_global):
                 ns_list = self.scheme.get('namespaceList', 'k8s')
             if not pod_list:  # use global resource file
                 pod_list = self.scheme.get('podList', 'k8s')
+
             pod_list = self._handle_resources_list(pod_list)
             ns_list = self._handle_resources_list(ns_list)
             peer_container = PeerContainer(ns_list, pod_list, config_name)
+            # look for service resources under 'pod_list' files
+            services = K8sServiceYamlParser.parse_service_resources(pod_list, peer_container)
```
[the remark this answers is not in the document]
tanyaveksler: It would create circular dependencies: K8sServiceYamlParser imports PeerContainer.
That's why I put service parser outside of PeerContainer, unlike ns and pod resource parser. Ideally, I think it would be more clean if ns resources and pod resources were parsed in K8sYamlParser or in their separate parsers.

network-config-analyzer/SchemeRunner.py
tanyaveksler: As far as I understand from the code, ns resources and pod resources may be global (per scheme), or per network config. Since we read service resources from pod list, we must try to read them per config as well, when supplied.

network-config-analyzer/K8sService.py
tanyaveksler: Added

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Done

tests/fw_rules_tests/policies/test1-scheme.yaml
tanyaveksler: Done

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Done

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Done

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Removed port syntax check

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,153 @@
...
+            service.set_type(K8sService.ServiceType.NodePort)
+        elif service_type == 'LoadBalancer':
+            service.set_type(K8sService.ServiceType.LoadBalancer)
+        else:
+            service.set_type(K8sService.ServiceType.ClusterIP)  # the default type
+
+        selector = service_spec.get('selector')
+        if selector:
+            for key, val in selector.items():
+                self.check_label_key_syntax(key, selector)
```
adisos, line 63:
```suggestion
```
A few more syntax checks we can remove

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,153 @@
...
+                self.check_label_value_syntax(val, key, selector)
```
adisos, line 64:
```suggestion
```

adisos (changes requested):

network-config-analyzer/SchemeRunner.py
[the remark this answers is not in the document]
adisos: So the current situation is that every time right after creating a `PeerContainer` object, you must also write the following code:
```
services = K8sServiceYamlParser.parse_service_resources(pod_list, peer_container)
peer_container.set_services(services)
```
This is currently only in `SchemeRunner`, but would have to be added also for `nca.py` where `PeerContainer` objects are also created (for cli queries).
It creates some code duplication + error-prone, so maybe we should break the dependency of `K8sServiceYamlParser` on `PeerContainer`. 
It seems that the dependency is originated from having the service yaml parser to assign the target pods for the service, and then for checking the named ports of the target pods.
One option is to take this functionality outside to a separate function to be run later after parsing all the services by the `PeerContainer`.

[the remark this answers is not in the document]
zivnevo: I agree with @adisos - `PeerContainer` should call `K8sServiceYamlParser`. Services are a bit more complex compared to Deployments and ServiceAccounts, but we should still try to hide this complexity from the users of `PeerContainer`.

As to the dependency problem, there are two places where `K8sServiceYamlParser` depends on `PeerContainer`:
1. It needs to call `PeerContainer.locate_kube_config_file()` before calling `CmdlineRunner` - this is really an old design bug. There is no reason for this function to reside inside `PeerContainer` - it should move to `CmdlineRunner` and always get called by `CmdlineRunner.get_k8s_resources()`.
2. It needs to set the set of pods behind the service by calling `peer_container.get_peers_with_label()`. This can be done by `PeerContainer` itself after calling the parser, using the key-values stored in `K8sService.selector`. Not extremely elegant, but can work 🤫

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Done

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Done

tanyaveksler:
I removed the dependency of K8sServiceYamlParser on PeerContainer, as you suggested.

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -0,0 +1,144 @@
...
+                    target_port = port_id
+                name = port.get('name', '')
+                if not service.add_port(K8sService.ServicePort(port_id, target_port,
+                                                               port.get('protocol', 'TCP'), name)):
+                    self.warning(f'The port {name} is not unique in Service {service.name}. Ignoring the port')
+                else:
+                    if isinstance(target_port, str):
+                        # check if all pods include this named port, and remove those that don't
+                        pods_to_remove = Peer.PeerSet()
+                        for pod in service.target_pods:
```
adisos, line 80: `service.target_pods` are not yet assigned in this stage

network-config-analyzer/K8sYamlParser.py
```diff
@@ -0,0 +1,202 @@
...
+        key = requirement['key']
+        operator = requirement['operator']
+        self.check_label_key_syntax(key, requirement)
+        if operator in ['In', 'NotIn']:
+            values = requirement.get('values')
+            if not values:
+                self.syntax_error('A requirement with In/NotIn operator but without values', requirement)
+            action = self.FilterActionType.In if operator == 'In' else self.FilterActionType.NotIn
+            if namespace_selector:
+                return peer_container.get_namespace_pods_with_label(key, values, action)
```
adisos, line 126: I'm seeing a lot of IDE warnings such as `Cannot find reference 'get_namespace_pods_with_label' in 'PeerContainer.py'` , because we are missing the import `from PeerContainer import PeerContainer`.
Adding the import would create some circular dependencies, but I think we can change the service yaml parse to inherit from the generic yaml parser instead of k8s yaml parser, so that we can add this import here.

adisos (changes requested):
added a few comments.

network-config-analyzer/K8sYamlParser.py
tanyaveksler: Changed the inheritance and united K8sYamlParser back with K8sPolicyYamlParser

network-config-analyzer/K8sServiceYamlParser.py
tanyaveksler: Fixed

network-config-analyzer/SchemeRunner.py
```diff
@@ -10,6 +10,7 @@
 from OutputConfiguration import OutputConfiguration
 from PeerContainer import PeerContainer
 from GenericYamlParser import GenericYamlParser
+from K8sServiceYamlParser import K8sServiceYamlParser
```
adisos, line 13:
```suggestion
```

network-config-analyzer/SchemeRunner.py
```diff
@@ -87,6 +88,7 @@ def _add_config(self, config_entry, peer_container_global):
                 ns_list = self.scheme.get('namespaceList', 'k8s')
             if not pod_list:  # use global resource file
                 pod_list = self.scheme.get('podList', 'k8s')
+
```
adisos, line 91:
```suggestion
```

network-config-analyzer/K8sService.py
```diff
@@ -0,0 +1,76 @@
...
+        self.name = name
+        self.namespace = namespace
+        self.type = self.ServiceType.ClusterIP
+        self.selector = {}
+        self.ports = {}  # a map from service port name to ServicePort object
+        self.target_pods = Peer.PeerSet()
+
+    def __eq__(self, other):
+        if isinstance(other, K8sService):
+            return self.name == other.name
```
adisos, line 41:
```suggestion
            return self.name == other.name and self.namespace == other.namespace
```
services are identified by the combination of namespace + name

network-config-analyzer/PeerContainer.py
```diff
@@ -172,6 +146,47 @@ def _set_peer_list(self, peer_resources_list, config_name):
...
+                    pod_named_port = pod.named_ports.get(port.target_port)
+                    if not pod_named_port:
+                        print(f'Warning: The named port {port.target_port} referenced in Service {srv.name}'
+                              f' is not defined in the pod {pod}. Ignoring the pod')
+                        pods_to_remove.add(pod)
+                srv.target_pods -= pods_to_remove
+
+            self.services[srv.name] = srv
+
+    def get_service(self, srv_name):
```
adisos, line 176: it seems that `get_service `and `get_services `are not in use

adisos (changes requested):
just a few more small comments

adisos:
hi @tanyaveksler , I tried the following small unit test on the topology example that you added, and it seems that the target pods are not populated:
```
import unittest
from PeerContainer import PeerContainer

class TestPeerContainerServices(unittest.TestCase):
    def test_basic(self):
        ns_list = ['topology/']
        pod_list = ['topology/']
        config_name = 'default'
        peer_container = PeerContainer(ns_list, pod_list, config_name)
```

Maybe the issue is because `get_peers_with_label `is expecting a namespace of type `K8sNamespace`, but you send a namespace of type `str`

network-config-analyzer/PeerContainer.py
```diff
@@ -40,44 +30,28 @@ def __init__(self, ns_resources=None, peer_resources=None, config_name='global')
         """
         self.peer_set = PeerSet()
         self.namespaces = {}  # mapping from namespace name to the actual K8sNamespace object
+        self.services = {}  # mapping from service name to the actual K8sService object
```
adisos, line 33: service name is not necessarily unique, as we may have 2 different services with the same name but on a different namespace.

network-config-analyzer/SchemeRunner.py
tanyaveksler: Done

network-config-analyzer/K8sService.py
tanyaveksler: Got it. Done.

network-config-analyzer/PeerContainer.py
tanyaveksler: removed

network-config-analyzer/PeerContainer.py
tanyaveksler: Done

tanyaveksler:
> 
Right, I did not treat namespaces correctly. When changed str to K8sNamespace, the target pods are populated.

network-config-analyzer/PeerContainer.py
```diff
@@ -172,6 +145,36 @@ def _set_peer_list(self, peer_resources_list, config_name):
...
+        and for every service computes and populates its target pods.
+        :param list srv_list: list of service in K8sService format
+        :return: None
+        """
+        for srv in srv_list:
+            # check and update namespace
+            srv.namespace = self.get_namespace(srv.namespace.name)
+            # populate target ports
+            for key, val in srv.selector.items():
+                srv.target_pods |= self.get_peers_with_label(key, [val], GenericYamlParser.FilterActionType.In,
```
adisos, line 160: The pods with labels from the service selector should be 'AND'ed  together

adisos (changes requested):
added another comment

adisos (dismissed):

zivnevo (dismissed):
ok to merge

adisos (approved):

Issue #134, opened 2022-01-04
activate tests performance tracking in the current workflow


adisos:
Activate the new flag for performance comparison in the run_tests script (on the workflow .github/workflows/test-push.yml  , update run_all_tests with type general to run with flag --check_run_time).
Use workflow artifacts to save the runtime comparison results.

adisos:
Based on the discussion on https://github.com/IBM/network-config-analyzer/pull/154:

- [ ] Add some mechanism to enable auto-update expected run-time when new tests are added or existing tests are modified.

shireenf-ibm:
>     * [ ]  Is it possible to add some warning if there are gaps?

Hi,
As I saw, a warning will be printed to the log as an artifact creation is printing.
But ,although we don't have any performance issues currently and max factor (for tests other than git_resources, which I raised its expected runtime) was 2.3. 
In future we'll not be able to tell if there are performance issues unless we check manually the existence of the "failed_runtime_check" artifact, a thing that may be forgotten.
So for example, shouldn't we increase the runtime_check factor up to 2.5 or 3 (instead of 2 ) 
and then to "stop/fail" the workflow run in case the failed_runtime_check artifact is created?

zivnevo:
> > ```
> > * [ ]  Is it possible to add some warning if there are gaps?
> > ```
> 
> Hi, As I saw, a warning will be printed to the log as the message of the creation of the artifact is printed. But ,although we don't have any performance issues currently and max factor (for tests other than git_resources, which I raised its expected runtime) was 2.3. In future we'll not be able to tell if there are performance issues unless we check manually the existence of the "failed_runtime_check" artifact, a thing that may be forgotten. So for example, shouldn't we increase the runtime_check factor up to 2.5 or 3 (instead of 2 ) and then to "stop/fail" the workflow run in case the failed_runtime_check artifact is created?

I suggest two thresholds: one for reporting in the artifact and (a bigger) one for making the test fail.

PR #154, opened 2022-01-26
activate testing performance in the workflow

shireenf-ibm:
#134

### DIFF
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -47,6 +47,11 @@ jobs:
         env:
           GHE_TOKEN: ${{ github.token }}
         run: python tests/run_all_tests.py --type=general --check_run_time
+      - name: Upload failed run-time tests file
+        uses: actions/upload-artifact@v2
+        with:
+          name: failed-run-time-check-file
+          path: ./tests/tests_failed_runtime_check.csv
       - name: Run output comparison tests
         run:  python tests/run_all_tests.py --type=output
       - name: Run fw-rules assertions tests
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -47,11 +47,11 @@ jobs:
         env:
           GHE_TOKEN: ${{ github.token }}
         run: python tests/run_all_tests.py --type=general --check_run_time
-#      - name: Commit failed run time file
-#        uses: EndBug/add-and-commit@v8.0.1
-#        with:
-#          default_author: github_actions
-#          message: 'Adding Failed Run Time Check'
+      - name: Commit failed run time file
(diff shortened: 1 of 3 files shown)


adisos (commented):
Hi @shireenf-ibm , the workflow file looks good.
How did you update the expected runtime? Is it based on some run results of the workflow?
How do we plan to auto-update expected runtime when modifying existing tests or adding new tests?

shireenf-ibm:
> Hi @shireenf-ibm , the workflow file looks good. How did you update the expected runtime? Is it based on some run results of the workflow? How do we plan to auto-update expected runtime when modifying existing tests or adding new tests?

Hi Adi,
yes, I updated the runtime values based on masters' workflow run on the github server.(first the values were based on running the tests on my local ws, but i realized much of them differ from the server's ones) 

after finishing issue #131 
I am planning to update the function that compare the runtime on run_all_tests.py , 
for new tests: to check for each scheme/cmdline file (when running with type=general) if it doesn't exist in the expected runtime file, we can add it with the current time 

but:
for modified tests: it will be a bit complicated , because we can't tell in code if the test file was modified since last run (updating time  if it differs may cause confusions, since on my local ws runtime may differ for many tests).
I am not very familiar with github actions, 
but if there is a way to tell if the file has changed and update its run time accordingly, 
or if there is a way to read the artifact that is created by this branch's action, and based on it to update the run time on the expected_runtime file  , this may help

adisos:
> > Hi @shireenf-ibm , the workflow file looks good. How did you update the expected runtime? Is it based on some run results of the workflow? How do we plan to auto-update expected runtime when modifying existing tests or adding new tests?
> 
> Hi Adi, yes, I updated the runtime values based on masters' workflow run on the github server.(first the values were based on running the tests on my local ws, but i realized much of them differ from the server's ones)
> 
> after finishing issue #131 I am planning to update the function that compare the runtime on run_all_tests.py , for new tests: to check for each scheme/cmdline file (when running with type=general) if it doesn't exist in the expected runtime file, we can add it with the current time
> 
> but: for modified tests: it will be a bit complicated , because we can't tell in code if the test file was modified since last run (updating time if it differs may cause confusions, since on my local ws runtime may differ for many tests). I am not very familiar with github actions, but if there is a way to tell if the file has changed and update its run time accordingly, or if there is a way to read the artifact that is created by this branch's action, and based on it to update the run time on the expected_runtime file , this may help

I assume we need a new workflow (to be run manually) for updating expected runtime (and commit the updates to the branch).
Since it is going to run in a workflow, I think we can do 'git diff', and send the diff results to the run_all_tests script, where we can consider only modified test files.
I will approve this PR and will keep issue #134 open, for this update to be added after closing issue #131.

adisos (approved):

PR #155, opened 2022-01-30
Bug in testing live cluster bug

tanyaveksler:
Added a delay after deploying resources in testing live cluster, to make sure all pods are up and running

### DIFF
--- a/tests/fw_rules_tests/podlist/kubernetes-manifests-new.yaml
+++ b/tests/fw_rules_tests/podlist/kubernetes-manifests-new.yaml
@@ -29,17 +29,20 @@ spec:
       labels:
         app: emailservice
     spec:
+      serviceAccountName: default
       terminationGracePeriodSeconds: 5
       containers:
       - name: server
-        image: gcr.io/google-samples/microservices-demo/emailservice:v0.1.3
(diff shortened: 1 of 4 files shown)


tanyaveksler:
Test_docker fails because gcc is missing - I remember you mentioned it in the project meeting, but I don't know how to workaround it.

zivnevo (dismissed):
ok to merge.
For now ignore the test_docker failure - I'll fix it later. You should be able to merge with this fail.

zivnevo:
I fixed the `test_docker` failure, Should be ok after you merge the latest changes from master.

zivnevo:
Also, maybe it's time to sync `tests/fw_rules_tests/podlist/kubernetes-manifests.yaml` with its [origin](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/release/kubernetes-manifests.yaml), as it got a little bit old.

For example, our version specifies using adservice version v0.1.3 - but there is no longer such version on [Google's registry](https://console.cloud.google.com/gcr/images/google-samples/global/microservices-demo/adservice), so the pod keeps failing.

zivnevo (approved):
ok to merge

PR #156, opened 2022-01-31
switch to python-slim

zivnevo:
Changes:
- Switched base image to `python-slim:3.8`, which is basically Debian Linux
- Using `apt-get` instead of `apk add` because this is the package manager in Debian
- Always downloading the latest version of `kubectl`
- Setting non-root user by simply setting the user to `9000`. No need for a new user-group and so.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,15 +3,15 @@
 # SPDX-License-Identifier: Apache2.0
 #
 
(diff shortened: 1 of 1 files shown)


zivnevo:
This should fix the failing `test-docker` job.

adisos (approved):
ok to merge

Issue #151, opened 2022-01-25
Support for GitHub raw content


zivnevo:
Currently, we only support urls like https://github.com/IBM/network-config-analyzer/blob/master/tests/k8s_testcases/example_policies/testcase1/testcase1-networkpolicy1.yaml

Users sometimes prefer to specify raw-content urls like https://raw.githubusercontent.com/IBM/network-config-analyzer/master/tests/k8s_testcases/example_policies/testcase1/testcase1-networkpolicy1.yaml
In that case, a simple http get can get us the file content - no need to go through GitHub API (but we can if we want to).

For now, let's just support single files in this format, without supporting (recursive) directories.

PR #160, opened 2022-02-07
supporting Ghe urls with raw content

shireenf-ibm:
#151


zivnevo (approved):
Looks great.
I like this nice trick of using a tuple in `str.startswith()`.

PR #165, opened 2022-02-13
adding runtime of new/missing tests to the tests_expected_runtime file

shireenf-ibm:
#134


shireenf-ibm:
summarizing the final commit:

- .github/workflows/test-push.yml : uploading the run log of Run all-tests job as an artifact to be used in the new workflow
- tests/run_all_tests.py : a "cosmetic" commit , to avoid printing empty lines in the failed_check_runtime.csv.
a note : in first commit I committed a function in this file to add new tests to the expected_runtime file, but since, I have the same functionality in my new script (which will always work for any new test committed) and the function in this file works only if the user run with the "check_run_time" flag (a thing I can't ensure will happen), I decided to revert that commit.

-  tests/tests_expected_runtime.csv: manual changes, adding new tests that are already on master branch and increasing the expected runtime for this "git_resource" tests , to avoid performance issues in these cases.(from slack decision)
- .github/workflows/update-tests-expected-runtime.yml:

1. the new workflow, will first get a list of the "changed-tests" in the PR (changed tests maybe : new scheme files (A), modified scheme files/ modified cmdline_tests file (M) , deleted tests files (D) , or renamed scheme/cmdline files (R) )
if there are no "changed_tests" , the workflow will stop running (nothing to do), else:
2. then will download the run_log from the new artifact in the test-push workflow (above).
3. Run the new script tests/update_expected_runtime.py to update the tests_expected_runtime.csv file 
4. remove the run_log from the repo and commit the tests_expected_runtime.csv 

Note: 
in the meanwhile I kept the workflow "on" section this way :
> on:
>   workflow_dispatch: # manual triggering
> 
> //  workflow_run:
> //    workflows: [test-push]
> //    types:
> //      - completed

when _workflow_dispatch_ is triggered manually only
and the _workflow_run_ (kept commented now) tells the workflow to run only after test-push workflow is completed.
However, both triggers work only from the master branch
so I couldn't test with them. (for my tests I used "pull_request" trigger with other combination to get the artifact)

- tests/update_expected_runtime.py : 
my new script , this script gets the "changed_tests" list from the previous job in the workflow.
and have an access to read the "run_log" which is downloaded from the artifact.
in this script :
**scheme files:**
new : added in new row to the tests_expected_runtime.csv  , with the actual runtime from the Run all tests workflow job (from the run_log)
modified : the script , updates the new runtime (from the run_log) in the relevant row of the modified test in the tests_expected_runtime.csv
deleted: the script delete the test's row in the tests_expected_runtime.csv

  **cmdline_file:**
  since we have multiple tests in one cmdline file, so for this file if  some queries were modified/ deleted (we can't tell which),
   the script will first delete all the cmdline tests from tests_expected_runtime.csv , then copy all cmdline_tests that were run from the run_log with there actual runtime

adisos (changes requested):
hi @shireenf-ibm, looks very good, a few comments:

- [ ] add another workflow to update all expected runtime of all tests ("reset") based on last run results
- [ ] scheme file updates: what if a scheme file is renamed?
- [ ] add some output for scripts in the action/workflow 
- [ ] two thresholds: one for reporting in the artifact and (a bigger) one for making the test fail.

shireenf-ibm:
> * scheme file updates: what if a scheme file is renamed?

when a file is renamed the `git diff` output is the file's new name.
so in this case , the "new file name" will be send as input to the script, and since it does not appear in the expected-run-time file, will be added  ( the script "thinks" as if it is a new added test), 
the only minus that the row (in the expected-run-time file) with the old name will not be removed cause the script has no indicator it does not exist anymore.
if we do not often rename test files , I suggest to keep it as it ("both rows").
however, Running The new (reset_expected_runtime) workflow will remove these "old" rows.

tests/run_all_tests.py
```diff
@@ -148,6 +148,8 @@ def _execute_run_time_compare(self, actual_run_time):
                     csv_writer.writerow(['test_name', 'expected_run_time (seconds)', 'actual_run_time (seconds)'])
                 csv_writer.writerow([self.test_name, expected_run_time, f'{actual_run_time:.2f}'])
             csv_file.close()
+            if actual_run_time > expected_run_time * 5:
```
adisos, line 151: make sure that expected run-time is not 0, so that we do not fail on some test that has not yet been added to expected runtime.

tests/update_expected_runtime.py
```diff
@@ -0,0 +1,152 @@
...
+
+def _add_cmdline_tests_to_file():
+    print('Updating cmdline tests with their new runtime')
+    lines = _get_run_log_summary_lines()
+    for line in lines:
+        if 'cmdline_' in line:
+            _add_test_to_expected_runtime_file(_get_test_name_from_line(line), _get_test_run_time_from_line())
+
+
+def _update_tests_runtime(modified_tests_list):
```
adisos, line 81: add some documentation about the current limitation of update when a test is renamed, due to output from 'git diff' in the workflow run

adisos (changes requested):
a few more comments

adisos (commented):
Hi @shireenf-ibm , looks very good.
One last comment: since you were not able to test the entire flow (including the commits that should be added following expected runtime updates), Can you prepare some test branch with some "test plan", so that you can open a PR (for testing) of the test branch right after we merge this PR?
You can change some tests in the test branch, so that we can test all kinds of expected runtime updates: deleted test, added test, modified test (with runtime greater/smaller than expected), renamed test.
Then after we merge, create a PR for the test branch, and then run the new workflow to see if a new commit with expected updates is added. Then, you can also test the new workflow that resets all expected runtime as well.

shireenf-ibm:
> Hi @shireenf-ibm , looks very good. One last comment: since you were not able to test the entire flow (including the commits that should be added following expected runtime updates), Can you prepare some test branch with some "test plan", so that you can open a PR (for testing) of the test branch right after we merge this PR? You can change some tests in the test branch, so that we can test all kinds of expected runtime updates: deleted test, added test, modified test (with runtime greater/smaller than expected), renamed test. Then after we merge, create a PR for the test branch, and then run the new workflow to see if a new commit with expected updates is added. Then, you can also test the new workflow that resets all expected runtime as well.

Sure, I think it is a great idea

adisos (approved):

PR #179, opened 2022-03-01
testing branch - to test new workflows

shireenf-ibm:


shireenf-ibm:
Hi @adisos ,
In previous commit , I see that the **update-tests-expected-runtime** workflow, behaved as desired, with all committed tests' changes.
The workflow pushed the updated tests_expected_runtime.csv successfully with all expected changes.

In next commits I'll revert the tests' changes and run the reset-tests-expected-runtime workflow

shireenf-ibm:
reset-tests-expected-runtime.yml workflow successfully completed as well

adisos:
Hi @shireenf-ibm , great - looks good.
Can you change the output of the file tests_expected_runtime.csv so that it will be sorted? This way it may be easier to view the diff and see what has been updated.
Also, there are 2 new tests from master branch that were not added to this file on master:
istio_testcases/example_policies/online_boutique/connectivity-scheme
k8s_testcases/example_policies/online_boutique/connectivity-scheme.yaml

shireenf-ibm:
> Hi @shireenf-ibm , great - looks good. Can you change the output of the file tests_expected_runtime.csv so that it will be sorted? This way it may be easier to view the diff and see what has been updated. Also, there are 2 new tests from master branch that were not added to this file on master: istio_testcases/example_policies/online_boutique/connectivity-scheme k8s_testcases/example_policies/online_boutique/connectivity-scheme.yaml

I see the new tests on the final file tests_expected_runtime.csv after running the "reset" workflow, lines 5 and 95.

the "reset" workflow copies the tests from the run log in same order that they appear there, for sorting I can copy them to a "dict" first then write to file, to do it?

the "update" workflow , adds new lines at the end of the file

adisos (commented):
Can you add some documentation in the script that updates the expected runtime, explaining that a change in the test k8s_cmdline_tests.yaml is going to "reset" all its queries expected runtime, and not only of the modified queries?

adisos (commented):
Looks good.
One more comment: What about the "git resources" tests (./k8s_testcases/example_policies/workload-resources-test/git-resource-test-scheme.yaml , k8s_cmdline_tests.yaml, query name: git_resources)?
We decided to manually increase the expected runtime for those tests, but this is going to be overridden with every reset/update of this file.
One option I can think of is to change the script that updates the expected runtime, so that for those tests it would create an expected runtime from actual runtime multiplied by some factor (1.5 or 2 maybe?)

adisos:
> Looks good. One more comment: What about the "git resources" tests (./k8s_testcases/example_policies/workload-resources-test/git-resource-test-scheme.yaml , k8s_cmdline_tests.yaml, query name: git_resources)? We decided to manually increase the expected runtime for those tests, but this is going to be overridden with every reset/update of this file. One option I can think of is to change the script that updates the expected runtime, so that for those tests it would create an expected runtime from actual runtime multiplied by some factor (1.5 or 2 maybe?)

Another option is to exclude these tests from runtime comparison.

adisos (approved):

adisos (approved):

PR #184, opened 2022-03-09
Issue 180

adisos:


zivnevo:
Does this remain a real issue after the switch to GlobalNetworkPolicy?
I'm afraid that if Istio one day introduce source-filtering using namespace-labels (rather than namespace names as it is now), things will break.

adisos:
> Does this remain a real issue after the switch to GlobalNetworkPolicy in baseline-rules-verifier? I'm afraid that if Istio one day introduce source-filtering using namespace-labels (rather than namespace names as it is now), things will break.

It remained a problem after the switch to Calico policy, because in Calico the default namespace is not added before parsing the policy (as opposed to Istio and k8s), so it resulted in a config of calico containing only 1 ns other than default (from the topology deployments), and another config of istio containing 2 ns: default and the other one (default being added only from parsing the policies, although policies were of the other ns).

I have pushed an update -- to be consistent with all three policy parsers (k8s, calico, istio): setting the namespace of a policy to default only when needed. (so that it is not being added to peer container in advance before doing any parsing).

network-config-analyzer/IstioPolicyYamlParser.py
```diff
@@ -598,13 +597,14 @@ def parse_policy(self):
...
-        # TODO: what if namespace is not specified in istio policy?
-        if 'namespace' in self.policy['metadata']:
-            self.namespace = self.peer_container.get_namespace(self.policy['metadata']['namespace'])
-
-        res_policy = IstioNetworkPolicy(self.policy['metadata']['name'], self.namespace)
+        if 'namespace' in metadata:
+            self.namespace = self.peer_container.get_namespace(metadata['namespace'])
+        else:
+            self.namespace = self.peer_container.get_namespace('default')
+        res_policy = IstioNetworkPolicy(metadata['name'], self.namespace)
```
zivnevo, line 607:
```suggestion
        self.namespace = self.peer_container.get_namespace(metadata.get('namespace', 'default'))
        res_policy = IstioNetworkPolicy(metadata['name'], self.namespace)
```

zivnevo (changes requested):
Shorter syntax

zivnevo (approved):
ok

Issue #153, opened 2022-01-26
Add new switch: resource_list


adisos:
All yaml resources (topology and policies) would be specified by this new switch (--resource_list)
On live cluster using this switch would override resources from the cluster.
For backward compatibility, preserve existing switches: ns_list, pod_list.
In future versions ns_list and pod_list will be depreciated

adisos:
Some aspects to consider:
- should manifests include policy files or only topology files?
- should support backward compatibility or not?
- how would be used for scheme files and for cli queries, with/without live cluster?

PR #188, opened 2022-03-20
issue_153_resource_list

shireenf-ibm:
_first commit.
Hi @adisos, @zivnevo 
The PR is not finished , this commit is to explain and show the changes idea, and get first review.

Since input lists may now be given only in resource_list (on input) and a priority should be if multiple lists are given (besides resource_list), the main idea is to handle and parse all input lists (resources) only once for topology and policies.
(implemented by adding the ResourcesHandler.py and ResourcesParser.py files)

In this commit, started to change "handling resources" for cmdline tests only (refactored nca.py) and by parsing the topology inputs: pods, namespaces and services (changef PeerContainer.py and K8sServiceYamlParser.py)

** the changes made in tests/run_all_tests.py only for successful testing - to be reverted


shireenf-ibm:
TODOs:
1. Parse the input resources in ResourcesParser for policies too and updating the ResourcesHandler accordingly 
2. fiishing some functionality in the resourcesParser
3. adding the "resourceList" option for scheme tests and handling input resources and lists .
4. support "resource list " flags to be additive (append action)
5. updating docs
6. adding more tests
7. update the RestServer to adapt with the changes in PeerContainer

shireenf-ibm:
some clarifications/questions:
1. nca.py:
- `manual_or_automatic.add_argument('--sanity', nargs='?', const='', type=_network_policies_valid_path,
                                     help='Network policies (file/dir/GHE url/cluster-type) for sanity checking')`
   for each --<query_name> arg, I've added the _nargs='?', const=''_ : so the np_list is optional after it (will be read from resource_list if not found ), if no query name is given , --sanity is the default.

- `parser.add_argument('--resource_list', '-r', type=_resource_list_valid_path,  # action='append',
                        help='Filesystem or GHE location of base network resources ')`
for resource_list and base_resource_list , I've added the type=_resource_list_valid_path , assuming the only allowed platform is k8s. (since it is the common one), Is this okay?

- however, I see that for some queries ( forbids, permits, and interferes ) the type is "str" and not "_network_policies_valid_path". 

- the resource_list flags will be additive (action=append), the ns_list and pod_list are additive too, but np_list is not additive. 
so, how should the parser behave if there are multiple network-policies paths appended to the resource_list flag (in case no specified np_list arg) ?

network-config-analyzer/nca.py
```diff
@@ -55,6 +57,13 @@ def _namespace_list_valid_path(path_location):
     return _valid_path(path_location, allow_ghe=True, allowed_platforms=['k8s'])
 
 
+def _resource_list_valid_path(path_location):
+    """
+    validation for paths of resources list in the command line
+    """
+    return _valid_path(path_location, allow_ghe=True, allowed_platforms=['k8s'])
```
adisos, line 64: `allowed_platforms `should contain all options `['k8s', 'calico', 'istio']` , because resource list may refer to all resource types. However we do need to think about allowed combinations of multiple values for resource list, where involving both paths and platform names.. Also, determine some dependencies, e.g. if resource list is istio, then policies should be for istio , pods and namespaces from k8s. If resource is calico, then policies should be for calico, pods from calico and namespaces from k8s.

network-config-analyzer/nca.py
```diff
@@ -171,24 +197,29 @@ def nca_main(argv=None):
...
-    manual_or_automatic.add_argument('--equiv', type=_network_policies_valid_path,
+    manual_or_automatic.add_argument('--equiv', type=_network_policies_valid_path, nargs='?', const='',
                                      help='Network policies (file/dir/GHE url/cluster-type) for equivalence checking')
-    manual_or_automatic.add_argument('--interferes', type=str, help='Network policies '
-                                     '(policy name/file/dir/GHE url/cluster-type) for interference checking')
-    manual_or_automatic.add_argument('--forbids', type=str, help='Network policies '
-                                     '(policy name/file/dir/GHE url/cluster-type) specifying forbidden connections')
-    manual_or_automatic.add_argument('--permits', type=str, help='Network policies '
-                                     '(policy name/file/dir/GHE url/cluster-type) specifying permitted connections')
+    manual_or_automatic.add_argument('--interferes', type=str, nargs='?', const='',
```
adisos, line 208: the types for interferes, forbids, permits, should also be network_policies_valid_path instead of str.

tests/k8s_cmdline_tests.yaml
```diff
@@ -156,3 +156,44 @@
...
+#- name: basic_sanity_unused_resource_pods_ns  # resource_list will be ignored cause there are pod_list and ns_list args
+#  args: >
+#    --sanity k8s_testcases/example_policies/testcase8/testcase8-networkpolicy1.yaml
+#    --resource_list k8s_testcases/example_podlist/bank_pod_list.json
+#    --resource_list k8s_testcases/example_podlist/bank_ns_list.json
+#    --ns_list k8s_testcases/example_podlist/ns_list.json
+#    --pod_list k8s_testcases/example_podlist/pods_list.json
+#  expected: 1
+
+- name: basic_sanity_used_resource_np  # the networkpolicy will be read from the resource_list
```
adisos, line 185: Can add a test where the ns_list and pod_list are also replaced by resource_list ? (is it supposed to work on this current version?)

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,42 @@
...
+        base_network_config = NetworkConfig(base_np_list, base_peer_container, [base_np_list])
+        return base_network_config
+
+    def get_network_config(self):
+        resources_parser = ResourcesParser(self.ns_list, self.pod_list,
+                                           self.resource_list, self.config_name)
+        peer_container = resources_parser.build_peer_container()
+        # TODO: parsing the network policies will be done by the ResourcesParser too
+        if self.np_list and self.resource_list:
+            print(f'the network policies will be taken only from {self.np_list}')
```
adisos, line 39: Add a "warning" prefix + maybe also specify what is ignored?

adisos (commented):
added a few initial comments

adisos:
> some clarifications/questions:
> 
>     1. nca.py:
> 
> 
>     * `manual_or_automatic.add_argument('--sanity', nargs='?', const='', type=_network_policies_valid_path, help='Network policies (file/dir/GHE url/cluster-type) for sanity checking')`
>       for each --<query_name> arg, I've added the _nargs='?', const=''_ : so the np_list is optional after it (will be read from resource_list if not found ), if no query name is given , --sanity is the default.
> 
>     * `parser.add_argument('--resource_list', '-r', type=_resource_list_valid_path,  # action='append', help='Filesystem or GHE location of base network resources ')`
>       for resource_list and base_resource_list , I've added the type=_resource_list_valid_path , assuming the only allowed platform is k8s. (since it is the common one), Is this okay?
> 
>     * however, I see that for some queries ( forbids, permits, and interferes ) the type is "str" and not "_network_policies_valid_path".
> 
>     * the resource_list flags will be additive (action=append), the ns_list and pod_list are additive too, but np_list is not additive.
>       so, how should the parser behave if there are multiple network-policies paths appended to the resource_list flag (in case no specified np_list arg) ?

hi @shireenf-ibm , I replied to some of your questions in the comments.
As for the last question: do you see a reason for not appending network policies from multiple paths in resource lists?

network-config-analyzer/nca.py
```diff
@@ -25,6 +25,8 @@ def _valid_path(path_location, allow_ghe=False, allowed_platforms=None):
     :rtype: str
     :raises: argparse.ArgumentTypeError when path is invalid
     """
+    if not path_location:
```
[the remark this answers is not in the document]
shireenf-ibm: I added this because the np_list after the query name can be an empty string now.
for example, we can write :
--equiv 
--resource_list <networkpolicy_path>
In this case np_list will be read as an empty string (from const='')
however this check is true only for np_lists so I may add it to the _network_policies_valid_path instead of here (_valid_path), to be more exact

tests/k8s_cmdline_tests.yaml
shireenf-ibm: yes , I will add it,
I couldn't add a test with separate paths (since action append is not supported yet because of handling the networkpolicies)
But I can attach the dir path of ns and pods to the resource_list , it should work

network-config-analyzer/nca.py
shireenf-ibm: thanks, I will start handling this

network-config-analyzer/ResourcesHandler.py
shireenf-ibm: Okay
I will add warning messages on what is ignored in the ResourcesParser (for topology and policies) - adding this to the TODOs list

shireenf-ibm:
Since this PR has a lot of refactoring and changes.
I would like to get a review, before implementing/ handling the "what to do on policies combinations"

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,182 @@
...
+        if np_list and np_list != ['']:
+            self._parse_resources_path(np_list, np_flag=True)
+            if resource_list:
+                print('Warning: policies will be taken from the provided networkPolicy list key. '
+                      'input resources provided with resource list key will be ignored when finding network policies')
+        elif resource_list:
+            print(f'Finding policies in {resource_list}')
+            self._parse_resources_path(resource_list, np_flag=True)
+
+    def _parse_resources_path(self, resource_list, np_flag=False, ns_flag=False, pod_flag=False):
```
adisos, line 135: Instead of using multiple input flags, can define an enumerated type of resource type to parse, and send a single value of this type.

network-config-analyzer/PoliciesFinder.py
```diff
@@ -0,0 +1,135 @@
...
+from IstioPolicyYamlParser import IstioPolicyYamlParser
+
+
+class PoliciesFinder:
+    """
+    This class is responsible for finding the network policies in the relevant input resources
+    The class contains several ways to build the set of policies (from cluster, from file-system, from GitHub).
+    """
+    def __init__(self):
+        self.policies_container = PoliciesContainer(policies={}, sorted_policies=[], profiles={}, allowed_labels=set())
```
adisos, line 26: `allowed_labels` seems not relevant to `policies_conatiner` (it is not a policy).
Maybe it should be computed in the level of NetworkConfig,  after collecting all policies.
It is basically a set of strings computed as the union of each policy's `referenced_labels`.
I also think there is an issue in NetworkConfig when using `append_policy_to_config` and not taking into account the added policy's  `referenced_labels` within the config's `allowed_labels`. 
For `clone_without_policies ` the network config should not contain any allowed labels, as it has no policies with no referenced labels.

network-config-analyzer/PoliciesFinder.py
```diff
@@ -0,0 +1,135 @@
...
+        self._parse_queue = deque()
+        self.type = NetworkConfig.ConfigType.Unknown
+        self.peer_container = None
+
+    def set_peer_container(self, peers):
+        self.peer_container = peers
```
adisos, line 32:
```suggestion
    def set_peer_container(self, peer_container):
        self.peer_container = peer_container
```
Also, add some documentation, including param types.

adisos (changes requested):
Hi @shireenf-ibm , looks good, added a few comments to start with.

adisos (commented):
a few more comments

network-config-analyzer/PoliciesFinder.py
```diff
@@ -0,0 +1,135 @@
...
+        elif policy_type == NetworkPolicy.PolicyType.CalicoProfile:
+            self._parse_queue.appendleft((policy_object, file_name, policy_type))  # profiles must be parsed first
+        else:
+            self._parse_queue.append((policy_object, file_name, policy_type))
+
+    def _add_policies_to_parse_queue(self, policy_list, file_name):
+        for policy in policy_list:
+            self.parse_yaml_code_for_policy(policy, file_name)
+
+    def _add_policies(self, doc, file_name, is_list=False):
```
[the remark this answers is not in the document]
shireenf-ibm: seems that we don't.
Checked on the master branch too (since it was copied as is)

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,223 @@
...
+                yaml_files = resource_scanner.get_yamls()
+                if not yaml_files:
+                    continue
+                for yaml_file in yaml_files:
+                    for res_code in yaml_file.data:
+                        if ResourceType.Namespaces in resource_flags:
+                            self.ns_finder.parse_yaml_code_for_ns(res_code)
+                        if ResourceType.Pods in resource_flags:
+                            self.pods_finder.namespaces_finder = self.ns_finder
+                            self.pods_finder.add_eps_from_yaml(res_code)
```
shireenf-ibm, line 199: In recent commit to the master branch, instead of sending the `res_code` to this function (`add_eps_from_yaml`), the `yaml_file.data` is sent.
I kept it with sending the res_code, since:
1. I think it works the same with res_code and yaml_file.data (I made all the changes in the function itself to match that commit's changes)
2. yaml_file.data is a generator , sending it as a parameter (to the functions) will cause tests with the new switch "resource_list" to fail (e.g for tests with resource_list that replaces ns_list and pod_list, the generator will find only namespaces)

So, I would like to know if keeping the `res_code` is okay for you. (is there any case that may fail if we don't send the `yaml_file.data`)

shireenf-ibm: From a short conversation with Ziv, this is good too

adisos (changes requested):
Added a few more comments.
Also,  consider adding a test with base_resource_list and resource_list.

network-config-analyzer/nca.py
```diff
@@ -38,6 +38,8 @@ def _network_policies_valid_path(path_location):
     """
     validation for paths of Network policies in the command line
     """
+    if not path_location:
```
adisos, line 41:
```suggestion
    if not path_location:  # when empty, expecting policies from resources list
```

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,233 @@
...
+                self._parse_resources_path(pod_resource, [ResourceType.Pods])
+
+        if len(self.pods_finder.peer_set) > 0 and len(self.ns_finder.namespaces) > 0:
+            return True, 0
+        elif len(self.pods_finder.peer_set) == 0 and len(self.ns_finder.namespaces) > 0:
+            return False, ResourceType.Namespaces
+        elif len(self.pods_finder.peer_set) > 0 and len(self.ns_finder.namespaces) == 0:
+            return False, ResourceType.Pods
+        else:
+            return False, 0
```
adisos, line 160: Why is it considered a failure if the length of pods or namespaces is 0?
Is the fallback in this case trying to get resources from live cluster?
Was this the behavior also in the previous version?
If the user specifies a path which as no pods resources at all, isn't it sufficient to report the size of the cluster topology (peer container)?

adisos (commented):
hi @shireenf-ibm , looks good, added a few comments

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,233 @@
...
+        else:
+            ns_resource = resource_list
+
+        if pod_list:
+            pod_resource = pod_list
+            if resource_list:
+                print('Warning: pods and services will be taken from the provided pod list key. '
+                      'input resources provided with resource list key will be ignored when finding pods and services')
+        else:
+            pod_resource = resource_list
```
adisos, line 139: Try to avoid this similar code repetition maybe with another separate method.

network-config-analyzer/ResourcesHandler.py
shireenf-ibm: before adding these "elif conditions" (from first commits), in case of "no namespaces/pods" are provided in the input resources, they'll be loaded from a k8s live cluster (this is also the behavior of old versions - always if ns_list/pod_list were not provided, then will be loaded from live cluster)

---------------------------------

these "elif" conditions were added **in case of "mixing global and private" resources in scheme files**.
for example, in scheme file, global resources may contain both pods and namespaces
but a specific network-config may override the ns_list --> in this case this method will find these private namespaces, and the fallback will use the global pods
in case no pods provided at all (neither global),  it will load them from k8s live cluster _( a fix is committed now regarding this)_

This keeps the origin behavior from old version for both cmdline and scheme tests

---------------------------------
so returning False in this method means "you can not build a peer container from the provided input resources" , with the resource type attached to the "False" result, the fallback knows to "fill" the missing parts

adisos: If I try to run 'nca --sanity -r empty_file.yaml' then it would try to get resources from live cluster, as opposed to the behavior on master branch when running 'nca --sanity empty_file.yaml --ns_list empty_file.yaml --pod_list empty_file.yaml' , where the output would be:
global: cluster has 0 unique endpoints, 0 namespaces
global: cluster has 0 unique endpoints, 0 namespaces
No NetworkPolicies in empty.json. Nothing to check sanity on.

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,256 @@
...
+        self.ns_finder = NamespacesFinder()
+        self.services_finder = ServicesFinder()
+
+    @staticmethod
+    def _determine_topology_resource(topology_list, resource_list, res_type):
+        res_name = 'namespaces' if res_type == ResourceType.Namespaces else 'pods and services'
+        if topology_list:
+            topology_resource = topology_list
+            if resource_list:
+                print(f'Warning: {res_name} will be taken from the provided namespace list key.'
```
adisos, line 123: The content of the warning says "will be taken from the provided **namespace**  list key" , but it's not always namespace

network-config-analyzer/ResourcesHandler.py
shireenf-ibm: the problem when providing a resource list , the "code" can not tell/predict what to search in it unless specific keys are provided.(specific keys that override it),

so if the resource list includes paths to namespaces and there are no pods at all (pod_list does not exist, and the resource_list does not include pods), it is hard to tell if the user intended to load pods from k8s or to run it with 0 pods.
 

I think I can tell the code that if "pod_list and ns_list" is provided "don't load from live cluster for empty files",
but what should I do when the objects are given with "resource_list"?

I would like to know your suggestion for these cases

shireenf-ibm:
Hi @adisos ,
following to the discussion on the input patterns that are  possible after adding the resource_list switch.
I have written all the patterns and started to run some examples on each pattern (examples with combinations of "good" and "bad" input paths).
 the main goal is to see how changes in this PR affected the behavior of loading the resources.

In the attached file, I included the patterns and examples of running cmdline queries with 1 config.
for each pattern , there is a table with some examples and the output of running each.

I added some notes in the table where the decision of loading the resources (especially pods) is "new".

[cmdline_1config_query_patterns_and_examples.xlsx](https://github.com/IBM/network-config-analyzer/files/8700245/cmdline_1config_query_patterns_and_examples.xlsx)

However, since  the ns_list and pod_list switches will depreciated (as described in the issue #153 ), I suggest that we focus on the patterns that don't include combinations of these switches with the resource_list.
particularly, we might "disable" or not encourage the use of the third and forth pattern in the attached file.


later, I'll attach similar files with patterns to run on cmdline_2_configs_queries or scheme files

shireenf-ibm:
attaching files that include the results of running all the examples I could  think of with most possible patterns.
[cmdline_1config_query_patterns_and_examples_updated.xlsx](https://github.com/IBM/network-config-analyzer/files/8760378/cmdline_1config_query_patterns_and_examples_updated.xlsx)
[cmdline_2_configs_query_patterns_and_examples.xlsx](https://github.com/IBM/network-config-analyzer/files/8760384/cmdline_2_configs_query_patterns_and_examples.xlsx)
[scheme_pattern_and_examples.xlsx](https://github.com/IBM/network-config-analyzer/files/8760386/scheme_pattern_and_examples.xlsx)

all examples on the attached files were tested on my final committed code changes

In addition, committed a new doc docs/CommonQueryPatterns.md with all desired patterns to be used and some references to examples
I also deleted tests with undesired patterns (involve resource_list with pod_list and ns_list) from previous commits

README.md
```diff
@@ -15,22 +15,28 @@ See an example scheme file [here](tests/k8s_testcases/example_policies/testcase1
...
+- `--connectivity [<NetworkPolicy set>]` \
 Get the list of allowed connections as firewall rules on the given set of NetworkPolicies
-- `--semantic_diff <NetworkPolicy set> [--base_np_list <NetworkPolicy set>]`\
+- `--semantic_diff [<NetworkPolicy set> --base_np_list <NetworkPolicy set>]`\
 Get the connectivity semantic difference as firewall rules between two sets of NetworkPolicy sets
 
+Note: The `<NetworkPolicy set>` may be provided instead within the --resource_list/--base_resource_list switches.\
+example: \
+  `--sanity`\
+  `--resorceList <NetworkPolicy set>`
```
adisos, line 38:
```suggestion
Note: The `<NetworkPolicy set>` may be provided instead within the `--resource_list/--base_resource_list` switches.\
Example:  `--sanity --resorceList <NetworkPolicy set>`
```

network-config-analyzer/ResourcesHandler.py
```diff
@@ -0,0 +1,292 @@
...
+
+        return peer_container
+
+    def _fill_empty_finder(self, res_type, resources_parser):
+        """
+        if res_type is ResourceType.Pods , then resources parser found only pods in the specific config,
+        use the global namespaces or load from k8s live cluster to build a peer container
+        else if res_type is ResourceType.Namespaces, then resources parser found only namespaces in the specific config,
+         use the global pods or load from k8s live cluster
+        """
```
adisos, line 101:
```suggestion
    def _fill_empty_finder(self, res_type, resources_parser):
        """
        :param ResourceType res_type: the topology resource type that was found from input resources
        :param ResourcesParser resources_parser: the current resources_parser object
        This function is called when one topology resource type is missing (the one different from input res_type).
        It updates resources_parser with relevant topology finder, either from global config or from live cluster:

        If res_type is ResourceType.Pods , then resources parser found only pods in the specific config,
        use the global namespaces or load from k8s live cluster to build a peer container
        If res_type is ResourceType.Namespaces, then resources parser found only namespaces in the specific config,
        use the global pods or load from k8s live cluster
        """
```

adisos (commented):
hi @shireenf-ibm , added a few more comments.

adisos (approved):

PR #189, opened 2022-03-23
compute_query_output fixes

shireenf-ibm:
I went through all "compute_query_output" and "exec" methods for all queries to make sure that the output is computed as expected (as described in the docs). 
For some queries, the behavior was as expected but minor fixes were done to print the full output results.
Forbids indeed didn't behave as expected for all cases, fixed

### DIFF
--- a/network-config-analyzer/NetworkConfigQuery.py
+++ b/network-config-analyzer/NetworkConfigQuery.py
@@ -132,7 +132,7 @@ class EmptinessQuery(NetworkConfigQuery):
                            full_explanation, res)
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, add_explanation=query_answer.bool_result)
 
 
 class VacuityQuery(NetworkConfigQuery):
@@ -237,7 +237,7 @@ class RedundancyQuery(NetworkConfigQuery):
         return QueryAnswer(False, 'No redundancy found in ' + self.config.name)
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, add_explanation=query_answer.bool_result)
 
 
 class SanityQuery(NetworkConfigQuery):
@@ -533,7 +533,7 @@ class ConnectivityMapQuery(NetworkConfigQuery):
         return res
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, only_explanation=query_answer.bool_result)
 
     @staticmethod
     def filter_istio_edge(peer2, conns):
@@ -938,8 +938,7 @@ class SemanticDiffQuery(TwoNetworkConfigsQuery):
         query_output = ''
         if self.output_config.outputFormat == 'txt':
             query_output += query_answer.output_result
-        if not query_answer.bool_result:
-            query_output += query_answer.output_explanation
+        query_output += query_answer.output_explanation
         return res, query_output
 
 
@@ -1047,7 +1046,7 @@ class TwoWayContainmentQuery(TwoNetworkConfigsQuery):
         explanation_not_contained_other_self = \
             contained_2_in_1.output_result + ':\n\t' + contained_2_in_1.output_explanation
         if contained_1_in_2.bool_result and contained_2_in_1.bool_result:
-            return QueryAnswer(bool_result=False,
+            return QueryAnswer(bool_result=True,
                                output_result=f'The two network configurations {self.name1} and {self.name2} '
                                              'are semantically equivalent.',
                                numerical_result=3)
@@ -1185,7 +1184,7 @@ class IntersectsQuery(TwoNetworkConfigsQuery):
                     return QueryAnswer(True, self.name2 + ' intersects with ' + self.name1, output_explanation)
 
         return QueryAnswer(False, f'The connections allowed by {self.name1}'
-                                  f' do not intersect the connections allowed by {self.name2}')
+                                  f' do not intersect the connections allowed by {self.name2}', numerical_result=1)
 
 
 class ForbidsQuery(TwoNetworkConfigsQuery):
@@ -1196,13 +1195,15 @@ class ForbidsQuery(TwoNetworkConfigsQuery):
                                       'No traffic is specified as forbidden.')
         return IntersectsQuery(self.config1, self.config2).exec(True)
 
-    def compute_query_output(self, query_answer, _):
-        if query_answer.bool_result:
-            query_output = f'{self.config2.name} does not forbid connections specified in {self.config1.name}:' \
+    def compute_query_output(self, query_answer, cmd_line_flag=False):
+        res = not query_answer.numerical_result if cmd_line_flag else query_answer.bool_result
+        query_output = query_answer.output_result + '\n'
+        if query_answer.numerical_result == 0:
+            query_output += f'{self.config2.name} does not forbid connections specified in {self.config1.name}:' \
                            f'{query_answer.output_explanation}'
         else:
-            query_output = f'{self.config2.name} forbids connections specified in {self.config1.name}'
-        return query_answer.bool_result, query_output
+            query_output += f'{self.config2.name} forbids connections specified in {self.config1.name}'
+        return res, query_output
 
 
 class AllCapturedQuery(NetworkConfigQuery):
--- a/network-config-analyzer/NetworkConfigQuery.py
+++ b/network-config-analyzer/NetworkConfigQuery.py
@@ -129,7 +129,7 @@ class EmptinessQuery(NetworkConfigQuery):
         full_explanation = '\n'.join(full_explanation_list)
         return QueryAnswer(res > 0,
                            'There are empty NetworkPolicies and/or empty ingress/egress rules in ' + self.config.name,
-                           full_explanation, res)
+                           '\n' + full_explanation, res)
 
     def compute_query_output(self, query_answer):
         return self.get_query_output(query_answer, add_explanation=query_answer.bool_result)
--- a/network-config-analyzer/NetworkConfigQuery.py
+++ b/network-config-analyzer/NetworkConfigQuery.py
@@ -1200,7 +1200,7 @@ class ForbidsQuery(TwoNetworkConfigsQuery):
         query_output = query_answer.output_result + '\n'
         if query_answer.numerical_result == 0:
             query_output += f'{self.config2.name} does not forbid connections specified in {self.config1.name}: ' \
-                        f'{query_answer.output_explanation}'
+                f'{query_answer.output_explanation}'
         else:
             query_output += f'{self.config2.name} forbids connections specified in {self.config1.name}'
         return res, query_output
--- a/network-config-analyzer/NetworkConfigQuery.py
+++ b/network-config-analyzer/NetworkConfigQuery.py
@@ -233,7 +233,7 @@ class RedundancyQuery(NetworkConfigQuery):
 
         if res > 0:
             output_explanation = '\n'.join(redundancies)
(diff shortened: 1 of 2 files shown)


network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1185,7 +1184,7 @@ def exec(self, only_captured=True):
                     return QueryAnswer(True, self.name2 + ' intersects with ' + self.name1, output_explanation)
 
         return QueryAnswer(False, f'The connections allowed by {self.name1}'
-                                  f' do not intersect the connections allowed by {self.name2}')
+                                  f' do not intersect the connections allowed by {self.name2}', numerical_result=1)
```
adisos, line 1187: Why is the numerical result of Intersects query set as 1 when bool result is False?
Would expect 1 for a bool result True..

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1196,13 +1195,15 @@ def exec(self):
                                       'No traffic is specified as forbidden.')
         return IntersectsQuery(self.config1, self.config2).exec(True)
 
-    def compute_query_output(self, query_answer, _):
-        if query_answer.bool_result:
-            query_output = f'{self.config2.name} does not forbid connections specified in {self.config1.name}:' \
-                           f'{query_answer.output_explanation}'
+    def compute_query_output(self, query_answer, cmd_line_flag=False):
+        res = not query_answer.numerical_result if cmd_line_flag else query_answer.bool_result
```
adisos, line 1199: This converts the result to bool even if required numerical result...
Can change to `query_answer.numerical_result` (without the `not`) and change the numerical result to be 1 in case of the bool result is True.

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1196,13 +1195,15 @@ def exec(self):
...
+        query_output = query_answer.output_result + '\n'
+        if query_answer.numerical_result == 0:
+            query_output += f'{self.config2.name} does not forbid connections specified in {self.config1.name}: ' \
```
adisos, line 1202: Output issue in forbids: when output is "The two configurations have different network topologies and thus are not comparable" or "There are no NetworkPolicies in the given forbids config. No traffic is specified as forbidden" ,
there is still additional appended output of "config2 does not forbid connections specified in config1" - though it is not relevant..

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1047,7 +1046,7 @@ def exec(self):
         explanation_not_contained_other_self = \
             contained_2_in_1.output_result + ':\n\t' + contained_2_in_1.output_explanation
         if contained_1_in_2.bool_result and contained_2_in_1.bool_result:
-            return QueryAnswer(bool_result=False,
+            return QueryAnswer(bool_result=True,
```
adisos, line 1049: Add a relevant test that will pass following this update and fail before this update.

adisos (commented):
Hi @shireenf-ibm , added a few comments.
Also, can you explain the other updates you have added that I did not refer to in comments - why they were required and are there any existing test outputs that demonstrate the change?

network-config-analyzer/NetworkConfigQuery.py
shireenf-ibm: - The numerical result is not used in other return statements in intersects , so it defaults to 0 and "meaningless" in this query's exec method.
but since intersects returns False (as bool result) in two cases : 
1. the configs are not comparable (from calling the "is_identical_topologies()"
2. when config1 do not intersect config2 
and we want to take only the 2nd case as really "forbidden" connection (later in Forbids.)
I used the numerical result as an indicator of the 2nd case.
if using "1" value is not good , I might use another value or
I might add a flag to the IntersectsQuery class to indicate a forbidden connection when calling the IntersectsQuery exec from Forbids.


I don't think adding numerical result = 1 when bool_result = True will help. because then we'll have to change the return value of the "is_identical_topologies()" which is called by other classes too.

Actually I thought that returning numerical_result = 0 with bool_result = True is good , cause exitting with 0 code (res)  means success

network-config-analyzer/NetworkConfigQuery.py
shireenf-ibm: Forbids calls the IntersectsQuery class exec method to compute the result so the same answer of the comment above is relevant here

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -533,7 +533,7 @@ def exec(self):
         return res
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, only_explanation=query_answer.bool_result)
```
shireenf-ibm, line 536: this is exactly the same as before, I added the flag name only to prevent confusion

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -132,7 +132,7 @@ def exec(self):
                            full_explanation, res)
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, add_explanation=query_answer.bool_result)
```
shireenf-ibm, line 135: before this change the query output always included only one part of the (output_result , output_explanation) , this ensures if the two parts exist in the queryAnswer so both will be printed in the query output message.
An example test: query name : with_emptiness , test file : tests/calico_testcases/example_policies/testcase19-profiles/testcase19-scheme.yaml

output before:

> Running query with_emptiness
Ingress rule no. 1 in NetworkPolicy np9-0-cnc-fe-between-namespaces-with-deny-all-profiles/cnc-fe/testcase19-9-1-specific-ingress-with-policies-based-apply-labels selects no source pods
NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels does not select any pods
Ingress rule no. 1 in NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels selects no destination pods
Egress rule no. 1 in NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels selects no source pods

output after:

> Running query with_emptiness
There are empty NetworkPolicies and/or empty ingress/egress rules in np9-0-cnc-fe-between-namespaces-with-deny-all-profiles
Ingress rule no. 1 in NetworkPolicy np9-0-cnc-fe-between-namespaces-with-deny-all-profiles/cnc-fe/testcase19-9-1-specific-ingress-with-policies-based-apply-labels selects no source pods
There are empty NetworkPolicies and/or empty ingress/egress rules in np11-global-without-label-to-apply
NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels does not select any pods
Ingress rule no. 1 in NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels selects no destination pods
Egress rule no. 1 in NetworkPolicy np11-global-without-label-to-apply/testcase19-11-global-open-within-cnc-kt-with-labels selects no source pods

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -237,7 +237,7 @@ def exec(self):
         return QueryAnswer(False, 'No redundancy found in ' + self.config.name)
 
     def compute_query_output(self, query_answer):
-        return self.get_query_output(query_answer, query_answer.bool_result)
+        return self.get_query_output(query_answer, add_explanation=query_answer.bool_result)
```
shireenf-ibm, line 240: same explanation of line 135 above.
an example:
query : redundantPoliciesInDenyFirstAlsoEgressDefaultAllow  file : tests/calico_testcases/example_policies/testcase11/testcase11-scheme.yaml

output before:

> Running query redundantPoliciesInDenyFirstAlsoEgressDefaultAllow
NetworkPolicy kube-system/testcase11-allow-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow
Ingress rule no. 1 in NetworkPolicy kube-system/testcase11-deny-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow
Egress rule no. 1 in NetworkPolicy kube-system/testcase11-default-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow

output after:

> Running query redundantPoliciesInDenyFirstAlsoEgressDefaultAllow
Redundancies found in np_denyFirstAlsoEgressDefaultAllow
NetworkPolicy kube-system/testcase11-allow-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow
Ingress rule no. 1 in NetworkPolicy kube-system/testcase11-deny-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow
Egress rule no. 1 in NetworkPolicy kube-system/testcase11-default-denyFirstAlsoEgress-defaultAllow is redundant in np_denyFirstAlsoEgressDefaultAllow

network-config-analyzer/NetworkConfigQuery.py
shireenf-ibm: - when having a test from a scheme file the important vlaue to be compared with the expected is the numerical result. so the bool result doesn't affect the final returned value in this case (meaningless)
- for cmdline tests the result is the "not bool" which means when there is a two way containment (True) we expect a "0" returned value (success) --> added the "two_way_containment_policies" test to the k8s_cmdline_tests.yaml file

the output of running the new query on both branches (this and master) will be :

>The two network configurations C:/Users/ShireenFalah/network-config-analyzer/tests/k8s_testcases/example_policies/testcase1/testcase1-networkpolicy1.yaml and C:/Users/ShireenFalah/network-config-analyzer/tests/k8s_testcases/example_policies/testcase1/testcase1-networkpolicy2.yaml are semantically equivalent.

The difference: 
on this branch it exit with Process finished with exit code 0
while on master branch Process finished with exit code 1

network-config-analyzer/NetworkConfigQuery.py
adisos: Now there is still additional appended output of "config2 forbids connections specified in config1"

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -1196,13 +1195,15 @@ def exec(self):
...
-        if query_answer.bool_result:
-            query_output = f'{self.config2.name} does not forbid connections specified in {self.config1.name}:' \
-                           f'{query_answer.output_explanation}'
+    def compute_query_output(self, query_answer, cmd_line_flag=False):
+        res = not query_answer.numerical_result if cmd_line_flag else query_answer.bool_result
+        query_output = query_answer.output_result + '\n'
+        if query_answer.bool_result and query_answer.numerical_result == 0:
+            query_output += f'{self.config2.name} does not forbid connections specified in {self.config1.name}: ' \
+                f'{query_answer.output_explanation}'
         else:
```
adisos, line 1204: I think this `else` should be replaced with `elif query_answer.numerical_result == 1` , to avoid adding this output when not relevant (as in my prev comment about not comparable/ no np cases.

network-config-analyzer/NetworkConfigQuery.py
adisos: ok, let's keep this as it is.

adisos (commented):
updated some comments replies + new comment

network-config-analyzer/NetworkConfigQuery.py
shireenf-ibm: True.
Actually since the only case the numerical_result = 1 is when there is no intersection between the existing policies.
the `and query_answer.numerical_result == 0 `from line 1201 is not needed
and the  `elif query_answer.numerical_result == 1` is required here

adisos (approved):
ok

Issue #205, opened 2022-04-24
connectivity map issue with ip-blocks


adisos:
1. It is not guaranteed that a policy's referenced ip-blocks are disjoint.
2. The method allowed_connections() for a policy works correct only if an input ip-block is either contained or disjoint from any referenced ip-block.
3. Thus connectivity map query has to be called with disjoint ip-blocks and not just referenced ip-blocks.

- Consider adding documentation for this method of allowed_connections() about this assumption.
- Consider also adding an assertion within this method.

PR #206, opened 2022-04-24
fix connectivity map disjoint ip blocks issue

adisos:
fixes issue #205 

- Added new query test within `tests/calico_testcases/example_policies/testcase16-nets/testcase16-scheme.yaml` , relevant for this issue.
- Moved the methods of disjoint ip-blocks into the IpBlock class, so that it is also used in ConnectivityMapQuery
- Fixed another issue of labels considered for fw-rules: adding the allowed labels as referenced labels from each policy through `add_policy` , so that for example labels are also added for configs generated from `clone_with_just_one_policy`.
- Added `.gitattributes` file to allow diffing dot files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.dot   diff=
+*.DOT   diff=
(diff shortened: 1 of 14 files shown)


zivnevo (approved):
Thanks. Code makes much more sense now

Issue #222, opened 2022-05-09
Parsing yaml policies - missing context at errors/warnings


adisos:
`GenericYamlParser `has methods `syntax_error `and `warning `which can get a `CommentedBase `object with context.
However, it seems this info is not available from the parsed policies.
For example, when running `nca --scheme tests\calico_testcases\example_policies\calico-demo\sanity1-scheme.yaml` , getting in output:
`Warning: selector selects no endpoints.`

On a much older version this used to work, with output as follows:
`tests\calico_testcases\example_policies\calico-demo\demo-sanity1-emptySelector.yaml:18:2: Warning: selector selects no endpoints.`

zivnevo:
There were changes to the API of `ruamel.yaml`.
In `GenericTreeScanner.py:54` we should now use `typ="rt"` rather than `typ="safe"`

PR #225, opened 2022-05-10
fix issue #222 + add context info for Istio parser warning about prin…

adisos:
…cipal with no match


zivnevo (changes requested):
`rt_load` is passed as a parameter to too may functions. I prefer it to be a data member of `GenericTreeScanner`, so it will only appear in the 3 constructors and the `get_scanner()` function.
(and the `_yield_yaml_file()` function will have to become non-static).

PR #227, opened 2022-05-10
 updating run_all_tests script

shireenf-ibm:
part1


adisos (commented):
Can you add some temporary change to some test so that it would fail the performance test, and so we can view the output log and make sure it works as expected?

adisos (commented):
great, thanks.
Let's revert the expected runtime change and then we can merge.

adisos (approved):

PR #234, opened 2022-05-24
Aggregation of TCP+UDP properties

shmfr:


network-config-analyzer/ConnectionSet.py
```diff
@@ -88,13 +94,65 @@ def _get_connections_representation(self, is_str):
...
 
+    @staticmethod
+    def _aggregate_connection_representation(protocols):
+        """
+        Aggregate shared properties of the protocols, for better human understanding.
+        :param dict protocols: a map from protocol number (1-255) to allowed properties
+        :return: dict protocols_not_aggregated: the rest of the protocol data that was not aggregated.
+        :return: str aggregation_results: a string of the aggregated representation
+        """
+        protocols_not_aggregated = protocols.copy()
```
zivnevo, line 105: Can we delay copying the data structure (may be expensive) until we are sure we'll need a copy?

zivnevo (changes requested):
I think we want to add testing for overlapping ranges. Either add a new test or add network configuration to one of the existing tests or modify an existing test.

zivnevo (approved):
Looks good :+1:

PR #238, opened 2022-06-02
deep copy fix

shireenf-ibm:
stdout fix as discribed  [here](https://github.com/np-guard/baseline-rules-verifier/pull/21)

peerContainer "statistics" message will be printed every time a Container is built or when a base config uses the same topology of the other ("global") config in the cmdline 2 configs queries

in addition, when a config uses  the global topology , it will use its "deep copy" instead of regular copy. I think this will be better, in cases any config make changes (adds) any peers/namespaces to the container

### DIFF
--- a/network-config-analyzer/ResourcesHandler.py
+++ b/network-config-analyzer/ResourcesHandler.py
@@ -27,7 +27,6 @@ class ResourcesHandler:
         self.global_peer_container = None
         self.global_pods_finder = None
         self.global_ns_finder = None
-        self.cmdline_two_configs_query = False
 
     def set_global_peer_container(self, global_ns_list, global_pod_list, global_resource_list):
         """
@@ -61,9 +60,6 @@ class ResourcesHandler:
         # build peer container
         peer_container = \
             self._set_config_peer_container(ns_list, pod_list, resource_list, config_name, save_flag, resources_parser)
-        if save_flag:
-            self.cmdline_two_configs_query = True  # this will indicate that if we use the global peer container for
-            # base topology too, still print its stats for both
 
         # parse for policies
         cfg = resources_parser.parse_lists_for_policies(np_list, resource_list, peer_container)
@@ -82,16 +78,12 @@ class ResourcesHandler:
             if res_type:
                 self._fill_empty_finder(res_type, resources_parser)
             peer_container = resources_parser.build_peer_container(config_name)
-            self._print_peer_container_stats(config_name, peer_container)
         elif self.global_peer_container:  # no specific peer container, use the global one if exists
             peer_container = copy.deepcopy(self.global_peer_container)
-            if self.cmdline_two_configs_query:
-                self._print_peer_container_stats(config_name, peer_container)
         else:  # the specific networkConfig has no topology input resources (not private, neither global)
             print('loading topology objects from k8s live cluster')
             resources_parser.load_resources_from_k8s_live_cluster([ResourceType.Namespaces, ResourceType.Pods])
             peer_container = resources_parser.build_peer_container(config_name)
-            self._print_peer_container_stats(config_name, peer_container)
 
         if save_flag:  # if called from scheme with global topology or cmdline with 2 configs query
             self.global_peer_container = peer_container
@@ -99,11 +91,6 @@ class ResourcesHandler:
             self.global_pods_finder = resources_parser.pods_finder
 
         return peer_container
-
-    @staticmethod
-    def _print_peer_container_stats(config_name, peer_container):
-        print(f'{config_name}: cluster has {len(peer_container.peer_set)} unique endpoints, '
-              f'{len(peer_container.namespaces)} namespaces')
 
     def _fill_empty_finder(self, res_type, resources_parser):
         """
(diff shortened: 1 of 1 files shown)


shireenf-ibm:
> stdout fix as discribed [here](https://github.com/np-guard/baseline-rules-verifier/pull/21)
> 
> peerContainer "statistics" message will be printed every time a Container is built or when a base config uses the same topology of the other ("global") config in the cmdline 2 configs queries
> 
> in addition, when a config uses the global topology , it will use its "deep copy" instead of regular copy. I think this will be better, in cases any config make changes (adds) any peers/namespaces to the container

I only kept the "deep copy" .
since the old version of nca , built a totally new peer container re-reading the resources in case of base config using the same topology of global config.

adisos:
Hi @shireenf-ibm , do we need to preserve this from the old version (the two separate copies of PeerContainer), or can we apply this optimization of using the same instance of the global peer container when the network config uses the global topology? Try to check where/if it is possible to change the peer container in the flow of the run, and whether the current implementation could cause an issue.

shireenf-ibm:
> Hi @shireenf-ibm , do we need to preserve this from the old version (the two separate copies of PeerContainer), or can we apply this optimization of using the same instance of the global peer container when the network config uses the global topology? Try to check where/if it is possible to change the peer container in the flow of the run, and whether the current implementation could cause an issue.

Hi, the only 2 changes that may be done to the peerContainer are when parsing the resources for policies (in the PoliciesFinder).
1. the first is adding the policy's namespace to the peer_container.namespaces if it doesn't exist there. - I don't think this may cause any issue
2. the second is clearing extra label pods - I don't see that this may cause an issue, since it is called before adding any policies  (also inn the old version). But as I see in the old version the deep copy was used because of it (in old SchemeRunner.py) - so I would like to know what do you think about it

all other methods that are used by NetworkConfig and relevant files in the PeerContainer are "get" methods, so are safe.

adisos:
> > Hi @shireenf-ibm , do we need to preserve this from the old version (the two separate copies of PeerContainer), or can we apply this optimization of using the same instance of the global peer container when the network config uses the global topology? Try to check where/if it is possible to change the peer container in the flow of the run, and whether the current implementation could cause an issue.
> 
> Hi, the only 2 changes that may be done to the peerContainer are when parsing the resources for policies (in the PoliciesFinder).
> 
>     1. the first is adding the policy's namespace to the peer_container.namespaces if it doesn't exist there. - I don't think this may cause any issue
> 
>     2. the second is clearing extra label pods - I don't see that this may cause an issue, since it is called before adding any policies  (also inn the old version). But as I see in the old version the deep copy was used because of it (in old SchemeRunner.py) - so I would like to know what do you think about it
> 
> 
> all other methods that are used by NetworkConfig and relevant files in the PeerContainer are "get" methods, so are safe.

About the second change that you mentioned - how do you infer that deep copy was used because of it?

shireenf-ibm:
> About the second change that you mentioned - how do you infer that deep copy was used because of it?

according to a comment msg that was written above the call of deepcopy.
quoting from the old version of the `SchemeRunner.py`: 

>    `#` deepcopy is required since NetworkConfig's constructor may change peer_container
            peer_container = copy.deepcopy(peer_container_global)

where the old version of the NetworkConfig's c'tr , used only ` peer_container.clear_pods_extra_labels()` (as ref to peerContainer's class) . 
following the old` __init__` of the NetworkConfig:

>         def __init__(self, name, peer_container, entry_list=None, config_type=None, buffer=None):
>           self.name = name
>           self.peer_container = peer_container
>           self._parse_queue = deque()  # This deque makes sure Profiles get parsed first (because of 'labelToApply')
>           self.policies = {}
>           self.sorted_policies = []
>           self.profiles = {}
>           self.referenced_ip_blocks = None
>           self.type = config_type or NetworkConfig.ConfigType.Unknown
>           self.allowed_labels = set()
>           peer_container.clear_pods_extra_labels()
>           if buffer is not None:
>               self._add_policies(buffer, 'buffer', True)
>           else:  # if entry_list is not None:
>               for entry in entry_list or []:
>                   self.add_policies_from_entry(entry)
>           self._parse_policies_in_parse_queue()
> 

since the call was used before parsing for policies, it was moved to the c'tr of PoliciesFinder.py

adisos:
> > About the second change that you mentioned - how do you infer that deep copy was used because of it?
> 
> according to a comment msg that was written above the call of deepcopy. quoting from the old version of the `SchemeRunner.py`:
> 
> > `#` deepcopy is required since NetworkConfig's constructor may change peer_container
> > peer_container = copy.deepcopy(peer_container_global)
> 
> where the old version of the NetworkConfig's c'tr , used only ` peer_container.clear_pods_extra_labels()` (as ref to peerContainer's class) . following the old` __init__` of the NetworkConfig:
> 
> > ```
> >     def __init__(self, name, peer_container, entry_list=None, config_type=None, buffer=None):
> >       self.name = name
> >       self.peer_container = peer_container
> >       self._parse_queue = deque()  # This deque makes sure Profiles get parsed first (because of 'labelToApply')
> >       self.policies = {}
> >       self.sorted_policies = []
> >       self.profiles = {}
> >       self.referenced_ip_blocks = None
> >       self.type = config_type or NetworkConfig.ConfigType.Unknown
> >       self.allowed_labels = set()
> >       peer_container.clear_pods_extra_labels()
> >       if buffer is not None:
> >           self._add_policies(buffer, 'buffer', True)
> >       else:  # if entry_list is not None:
> >           for entry in entry_list or []:
> >               self.add_policies_from_entry(entry)
> >       self._parse_policies_in_parse_queue()
> > ```
> 
> since the call was used before parsing for policies, it was moved to the c'tr of PoliciesFinder.py

ok, in the old c'tor of networkConfig it seems that also parsing the policies may affect the peerContainer (adding namespaces from policies).
Let's be consistent with the deepcopy() from the older version (some queries refer to gaps in topologies from peerContainer, so we probably don't want that a network config with some policies would affect the topology's namespaces of another network config that is based on global topology).
Can you add in your suggested change a comment explaining why deep copy is required? (similar to the comment from the old version, just adjusted to the current implementation).

adisos (approved):

Issue #240, opened 2022-06-06
Crash when trying to read endpoints from calico live cluster


zivnevo:
Running `nca.py --connectivity calico --pod_list calico` I get:
```
Traceback (most recent call last):
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\nca.py", line 278, in <module>
    sys.exit(nca_main())
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\nca.py", line 270, in nca_main
    ret_val = run_args(args)
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\nca.py", line 172, in run_args
    network_config = resources_handler.get_network_config(np_list, ns_list, pod_list, resource_list,
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\ResourcesHandler.py", line 62, in get_network_config
    self._set_config_peer_container(ns_list, pod_list, resource_list, config_name, save_flag, resources_parser)
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\ResourcesHandler.py", line 76, in _set_config_peer_container
    success, res_type = resources_parser.parse_lists_for_topology(ns_list, pod_list, resource_list)
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\ResourcesHandler.py", line 171, in parse_lists_for_topology
    self._parse_resources_path(pod_resource, [ResourceType.Pods])
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\ResourcesHandler.py", line 242, in _parse_resources_path
    self._handle_calico_inputs(resource_flags)
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\ResourcesHandler.py", line 282, in _handle_calico_inputs
    self.pods_finder.load_peer_from_calico_resource()
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\TopologyObjectsFinder.py", line 34, in load_peer_from_calico_resource
    self.add_eps_from_yaml(peer_code)
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\TopologyObjectsFinder.py", line 59, in add_eps_from_yaml
    self.add_eps_from_yaml(endpoint, kind[:-4])
  File "C:\nevo\projects\network-config-analyzer\network-config-analyzer\TopologyObjectsFinder.py", line 65, in add_eps_from_yaml
    self._add_wep_from_yaml(yaml_obj)
[...]

PR #242, opened 2022-06-07
assign ns_finder to pods_finder before loading calico peers

shireenf-ibm:
#240 
Hi @zivnevo , 
unfortunately, I couldn't run "calicoctl" successfully on my local workspace yet.
I think the bug occurred because of a "missing" assignment of namespaces_finder before loading calico pods. (fixed in this commit)
Can you please re-run the cmdline with this branch to check if the bug fixed? 
thanks


zivnevo (approved):
Looks good. I checked and there was no crash

Issue #243, opened 2022-06-07
crash of NoneType


shireenf-ibm:
TypeError: NoneType object is not iterable in _add_pod_from_workload_yaml
occured when running nca on yamls from helm-charts

zivnevo:
![image](https://user-images.githubusercontent.com/79099626/172382340-2d5df702-14df-4d13-9780-2175ac616743.png)

PR #244, opened 2022-06-07
iterating on NoneType

shireenf-ibm:
#243
The bug occurred when a container has an empty "ports" field so code read it as None instead of empty list  
(A fix is committed in `_add_pod_from_workload_yaml `(where it happened) and some other similar functions)

image


network-config-analyzer/TopologyObjectsFinder.py
```diff
@@ -102,8 +102,10 @@ def _add_pod_from_yaml(self, pod_object):
 
         containers = pod_object['spec'].get('containers', {})
         for container in containers:
-            for port in container.get('ports', {}):
-                pod.add_named_port(port.get('name'), port.get('containerPort'), port.get('protocol', 'TCP'))
+            ports = container.get('ports', {})
+            if ports:
+                for port in ports:
+                    pod.add_named_port(port.get('name'), port.get('containerPort'), port.get('protocol', 'TCP'))
```
zivnevo, line 108:
```suggestion
            for port in container.get('ports') or  []:
                pod.add_named_port(port.get('name'), port.get('containerPort'), port.get('protocol', 'TCP'))
```
Avoiding extra indentation. Please also apply to the rest of the changes

zivnevo (changes requested):

zivnevo (approved):

Issue #10, opened 2021-04-22
Support Calico GlobalNetworkPolicies with applyOnForward=True


zivnevo:
See https://docs.projectcalico.org/reference/resources/globalnetworkpolicy and https://docs.projectcalico.org/reference/host-endpoints/forwarded

zivn:
We should also support policies with preDNAT=true and doNotTrack=True

zivnevo:
Maybe for now, allow setting these flags to `True` but issue a warning on analysis being an over-approximation...

zivnevo:
Moved to new repo. See [here](https://github.com/np-guard/network-config-analyzer/issues/42)

PR #249, opened 2022-06-08
issue #10: allow specifying Calico's unsupported keys with a warning …

adisos:
…of analysis being an over-approximation


zivnevo (approved):
Thanks, looks good.

PR #272, opened 2022-07-05
adding indication of nca err for forbids/permits query

shireenf-ibm:
in some cases the query result is 1 not because it was executed and failed but from other reasons (such as, no networkpolicies in the config).
since baseline-rules-verifier uses `forbids `and `permits` queries and needs to know if the query was executed and violation is found or it didn't even execute. I committed here some changes to help identify the reason of query_res = 1 (if it was nca_err / violated)
Currently, changes are meaningful only for cmdline nca run and only for forbids and permits queries


adisos (commented):
Hi @shireenf-ibm ,

- [x] Can we avoid changing the type of the return value (it was changed from integer to tuple).
I think we can update the exit code meaning (for instance if its 8 or above this would indicate the query did not run, and the values for query res and output comparison res should be ignored).

- [x] Can we update this new indication result for all queries? What about scheme files?

shireenf-ibm:
a new flag : `query_not_executed`  was added for the query answer and is computed for all queries (the cases when this flag turns on are described in `docs/CmdLineQueriesResults.md`).
For cmd_line queries this flag will be returned and considered as the 3rd bit of the nca result (description in `nca.py `and `README.md`)
For the SchemeRunner, the new returned value is not used yet.
If we want to use it we may add it to the res value and consider this in the computation of the "expected" field's value.
or we may add a new attribute for each query in the scheme files. Alternatively we may print a warning/ pay attention message in case some queries were not checked

adisos:
> a new flag : `query_not_executed` was added for the query answer and is computed for all queries (the cases when this flag turns on are described in `docs/CmdLineQueriesResults.md`). For cmd_line queries this flag will be returned and considered as the 3rd bit of the nca result (description in `nca.py `and `README.md`) For the SchemeRunner, the new returned value is not used yet. If we want to use it we may add it to the res value and consider this in the computation of the "expected" field's value. or we may add a new attribute for each query in the scheme files. Alternatively we may print a warning/ pay attention message in case some queries were not checked

Let's add a new attribute for each query in scheme files (expected result for query_not_executed). Also consider adding a warning message in case that some queries were not executed.

docs/CmdLineQueriesResults.md
```diff
@@ -11,3 +11,8 @@ The result of each command-line query may be 0 or 1 as followed:
 | --connectivity | ConnectivityMap | the result of this query is always 0 | None |
 | --semantic_diff | SemanticDiff | both sets of NetworkPolicies are semantically equivalent | the given sets of NetworkPolicies are not semantically equivalent |
 
+## A query will not be executed when:
+1. The config/s type is not relevant for performing the query.
+2. An input config does not include NetworkPolicies.
+3. The input configs are not comparable.
+4. The input configs of equivalence/semantic-diff query are identical.
```
adisos, line 18: Are these cases relevant for only certain queries? which?

README.md
```diff
@@ -97,15 +97,17 @@ The arguments to `--resource_list` and to `--base_resource_list` should be one o
 For more information on command-line switches combinations, see [Common Query Patterns](docs/CommonQueryPatterns.md#cmdline-queries)
 
 #### Exit Code Meaning:
-The exit value of running a command-line without a scheme is the combination of two factors:
-1. The result of running the query (0/1) as specified [here](docs/CmdLineQueriesResults.md)
-2. The result of comparing the query output with the expected output file contents (if given)
+The exit value of running a command-line without a scheme is the combination of three factors:
+1. The query was not executed because of one of the reasons listed [here](docs/CmdLineQueriesResults.md#a-query-will-not-be-executed-when). If this is true, then other factors are ignored.
```
adisos, line 101: Move this to be the lase item in the list.

docs/SchemeFileFormat.md
```diff
@@ -65,6 +65,7 @@ Each query object instructs the tool to run a specific check on one or more sets
 |connectivityMap|Reports a summary of the allowed connections in the cluster|list of [config set](#configsets) names| 
 |expected|The expected sum of returned results by all sub-queries in this query (a warning is issued on mismatch)|integer|
 |expectedOutput|The file path of the expected output of this query (for connectivityMap or semanticDiff queries) |string|
+|expectedUnchecked|The number of input configs/config pairs that the query is not expected to be run on. Reasons for not checking the configs are listed [here](CmdLineQueriesResults.md#a-query-will-not-be-executed-when) |integer|
```
adisos, line 68: Some mix in the terminology - sometime referring to "unchecked queries" and sometimes to "queries not executed".
Let's be consistent and replace "unchecked" with "not executed" (code + documentation).

network-config-analyzer/nca.py
```diff
@@ -87,10 +87,10 @@ def _execute_single_config_query(query_name, network_config, output_config, expe
     :param str expected_output: a file path to the expected output
     :rtype: int
     """
-    res, comparing_err = NetworkConfigQueryRunner(query_name, [network_config], expected_output,
-                                                  output_config).run_query()
+    res, comparing_err, query_not_executed = NetworkConfigQueryRunner(query_name, [network_config], expected_output,
+                                                                      output_config).run_query()
     expected_res_bit = res > 0
-    return 2 * comparing_err + expected_res_bit
+    return 4 * query_not_executed + 2 * comparing_err + expected_res_bit
```
adisos, line 93: This computation repeats twice - maybe we can have it in a separate function?

adisos (commented):
Hi @shireenf-ibm , looks good.  Added a few comments.

docs/CmdLineQueriesResults.md
shireenf-ibm: my intention here is to describe general cases that may occur.
for example by the first case `The config/s type is not relevant for performing the query.` I mean the queries that do not run on Ingress configurations.
2. corresponds to queries with the `if not self.config` check
3. is for 2 configs queries that call the `is_identical_topologies` def
4. describes a case that is relevant only for equivalence and semantic diff as written

adisos (commented):
Hi @shireenf-ibm , looks good. I think we should revisit the cases where we determine that the query was not executed.
If the configurations have different topologies and thus are not comparable, then it makes sense to say that the query was not executed (for example IntersectsQuery).
But for example if the two configurations are identical then we have a valid result for the query (StrongEquivalence for example), so I think we should not say that the query is not executed on this case.

shireenf-ibm:
> Hi @shireenf-ibm , looks good. I think we should revisit the cases where we determine that the query was not executed. If the configurations have different topologies and thus are not comparable, then it makes sense to say that the query was not executed (for example IntersectsQuery). But for example if the two configurations are identical then we have a valid result for the query (StrongEquivalence for example), so I think we should not say that the query is not executed on this case.

I see, so is it better to eliminate the 4th case [here](https://github.com/IBM/network-config-analyzer/blob/identify_nca_errors_for_baseline_rules_verifier/docs/CmdLineQueriesResults.md#a-query-will-not-be-executed-when):

>  
1. The config/s type is not relevant for performing the query.
2. An input config does not include NetworkPolicies.
3. The input configs are not comparable.
4. The input configs of equivalence/semantic-diff query are identical.

and not turn on the `query_not_executed` flag when configs are identical ?

network-config-analyzer/NetworkConfigQuery.py
```diff
@@ -447,7 +448,8 @@ def redundant_policy_text(self, policy):
 
     def exec(self):  # noqa: C901
         if not self.config:
-            return QueryAnswer(False, f'No NetworkPolicies in {self.config.name}. Nothing to check sanity on.', '', 1)
+            return QueryAnswer(False, f'No NetworkPolicies in {self.config.name}. Nothing to check sanity on.', '',
+                               query_not_executed=True)
```
shireenf-ibm, line 452: Hi @adisos,
when a config has no policies in sanity check. the current query output is 
> No NetworkPolicies in {self.config.name}. Nothing to check sanity on.', '',
so , I turned on the `query_not_executed`.
If it shouldn't be turned on, wouldn't be better to change the query output?

shireenf-ibm:
> > > Hi @shireenf-ibm , looks good. I think we should revisit the cases where we determine that the query was not executed. If the configurations have different topologies and thus are not comparable, then it makes sense to say that the query was not executed (for example IntersectsQuery). But for example if the two configurations are identical then we have a valid result for the query (StrongEquivalence for example), so I think we should not say that the query is not executed on this case.
> > 
> > 
> > I see, so is it better to eliminate the 4th case [here](https://github.com/IBM/network-config-analyzer/blob/identify_nca_errors_for_baseline_rules_verifier/docs/CmdLineQueriesResults.md#a-query-will-not-be-executed-when):
> > > 
> > 
> > 
> > ```
> > 1. The config/s type is not relevant for performing the query.
> > 
> > 2. An input config does not include NetworkPolicies.
> > 
> > 3. The input configs are not comparable.
> > 
> > 4. The input configs of equivalence/semantic-diff query are identical.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and not turn on the `query_not_executed` flag when configs are identical ?
> 
> Yes, and also I think you should check for each query separately what makes sense. Another example is sanity query -- if the config has no network policies, the result that there are no violations is valid, and we don't need to say that the query did not run on this case.

docs/CmdLineQueriesResults.md
```diff
@@ -11,3 +11,7 @@ The result of each command-line query may be 0 or 1 as followed:
 | --connectivity | ConnectivityMap | the result of this query is always 0 | None |
 | --semantic_diff | SemanticDiff | both sets of NetworkPolicies are semantically equivalent | the given sets of NetworkPolicies are not semantically equivalent |
 
+## A query will not be executed when:
+1. The config/s type is not relevant for performing the query.
+2. An input config does not include NetworkPolicies.
```
adisos, line 16: I think the second case is not accurate.. it's only on permits/forbids query when there given permit/forbid config is empty.

adisos (commented):
Looks good, added one more comment.

adisos (commented):
Hi @shireenf-ibm , thanks, looks good. Just added one more comment.

adisos (approved):

Issue #281, opened 2022-07-19
No report for failed performance test


zivnevo:
I had a run of `test-push` workflow where `istio-tests` failed because of a performance issue.
However, no artifact was generated with details of failed runtime checks.
(I did get a `calico-failed-run-time-check-file` artifact, but I don't think it's related)

You can see this run [here](https://github.com/IBM/network-config-analyzer/actions/runs/2691907769/attempts/1). Note that I reran this test to make sure this is not a transient issue, and the second run passed.

zivnevo:
I also don't get why the test has failed. The expected runtime for `istio_testcases/example_policies/bookinfo-demo/bookinfo-test-request-attrs-scheme.yaml` is 3.37 seconds and the actual runtime was 3.52 seconds.

zivnevo:
Also, I got a second runtime-check failure for the same test: `fw_rules_tests/policies/istio-bookinfo-test-request-attrs-scheme.yaml  `. We should try to understand why this happens (actual 3.5-4 seconds vs. expected 0.67 seconds)

PR #282, opened 2022-07-19
bug fix 

shireenf-ibm:
Signed-off-by: shireenf-ibm <shireenf@ibm.com>
#281 
all failed_runtime files were removed by all categories runs , thus only calico's file was kept (since its the longest job)

### DIFF
--- a/.github/workflows/test-push.yml
+++ b/.github/workflows/test-push.yml
@@ -55,6 +55,7 @@ jobs:
           name: k8s-log
           path: tests/k8s_log.txt
       - name: Upload k8s failed run-time tests file
+        if: ${{ always() }}
         uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8
         with:
           name: k8s-failed-run-time-check-file
@@ -76,6 +77,7 @@ jobs:
           name: calico-log
           path: tests/calico_log.txt
       - name: Upload calico failed run-time tests file
+        if: ${{ always() }}
         uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8
         with:
           name: calico-failed-run-time-check-file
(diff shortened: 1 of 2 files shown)


shireenf-ibm:
Hi @ziv, 
the bug was resolved.

regarding the `fw_rules_tests/policies/istio-bookinfo-test-request-attrs-scheme.yaml` test,
from a look on its part in the log, I see that all of its queries are finished in 0.05/0.06 seconds except one query (finished in 1.88 in the last commit)

> Query connectivity-istio-test-methods-paths-1 finished in   1.88 seconds

However, this query appears twice in the scheme file, once with txt output format (finished in 0.05) and the second with the yaml format (finished with- 1.88)
Locally , when I run the scheme file only with this query in its two formats  and without `expectedOutput` entry (to avoid comparing results )
I get the following:
txt format:
Query connectivity-istio-test-methods-paths-1 finished in   0.06 seconds
yaml format:
Query connectivity-istio-test-methods-paths-1 finished in   2.37 seconds

So I assume that writing the query output itself i.e in this case building the connectivity graph took too long for the yaml format 
Does that make sense?

zivnevo:
> Hi @ziv, the bug was resolved.
> 
> regarding the `fw_rules_tests/policies/istio-bookinfo-test-request-attrs-scheme.yaml` test, from a look on its part in the log, I see that all of its queries are finished in 0.05/0.06 seconds except one query (finished in 1.88 in the last commit)
> 
> > Query connectivity-istio-test-methods-paths-1 finished in   1.88 seconds
> 
> However, this query appears twice in the scheme file, once with txt output format (finished in 0.05) and the second with the yaml format (finished with- 1.88) Locally , when I run the scheme file only with this query in its two formats and without `expectedOutput` entry (to avoid comparing results ) I get the following: txt format: Query connectivity-istio-test-methods-paths-1 finished in 0.06 seconds yaml format: Query connectivity-istio-test-methods-paths-1 finished in 2.37 seconds
> 
> So I assume that writing the query output itself i.e in this case building the connectivity graph took too long for the yaml format Does that make sense?

I saw problems with another test. I really don't know why there is such a difference between json and yaml.

PR #286, opened 2022-07-28
Issue #275: Checking that a label value is a string

### LINKED ISSUE #275: Labels which are numbers are treated as integers and not as strings

zivnevo:
Fixes #275 
Signed-off-by: Ziv Nevo <nevo@il.ibm.com>


shireenf-ibm (approved):
Approved.
A val must be string.
In order to use value equal to "number" , it must be provided in a string form. i.e `app: '6'` (confirmed this in the live cluster and on nca)

Issue #298, opened 2022-08-09
Identical __str__ and __repr__ methods in CanonicalIntervalSet


shaigue:
In the [CanonicalIntervalSet](https://github.com/IBM/network-config-analyzer/blob/f87ef7ffef016c9961b549200826f22d6d993795/network-config-analyzer/CanonicalIntervalSet.py#L38-L58)  class, the `__str__` and `__repr__` methods are identical.
I suggest to refactor `__repr__` to call `__str__`.

```
    def __str__(self):
        if not self.interval_set:
            return "Empty"
        res = ''
        for interval in self.interval_set:
            res += str(interval.start)
            if interval.start != interval.end:
                res += '-' + str(interval.end)
            res += ','
        return res[0:-1]

    def __repr__(self):
        if not self.interval_set:
            return "Empty"
        res = ''
        for interval in self.interval_set:
            res += str(interval.start)
            if interval.start != interval.end:
                res += '-' + str(interval.end)
            res += ','
        return res[0:-1]
```

PR #299, opened 2022-08-09
refactored __repr__ to call __str__

shaigue:
#298 @adisos

### DIFF
--- a/network-config-analyzer/CanonicalIntervalSet.py
+++ b/network-config-analyzer/CanonicalIntervalSet.py
@@ -35,7 +35,7 @@ class CanonicalIntervalSet:
     def __iter__(self):
         return iter(self.interval_set)
 
-    def __str__(self):
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


zivnevo:
[Python documentation](https://docs.python.org/3/reference/datamodel.html?highlight=__str__#object.__str__) says that if `__repr__`  is defined, and `__str__` is not, the object will behave as though `__str__=__repr__`.
So maybe leave `__repr__` as-is and  just delete `__str__`

shaigue:
> [Python documentation](https://docs.python.org/3/reference/datamodel.html?highlight=__str__#object.__str__) says that if `__repr__` is defined, and `__str__` is not, the object will behave as though `__str__=__repr__`. So maybe leave `__repr__` as-is and just delete `__str__`

OK, did that @zivnevo

zivnevo (approved):
Thanks.
Please use `Squash and Merge` to merge the PR.

PR #318, opened 2022-08-25
Support Istio Sidecar

shireenf-ibm:
sidecar support, with priority calculated in allowed_connections stage


network-config-analyzer/IstioSidecar.py
```diff
@@ -0,0 +1,122 @@
...
+        :param IstioSidecarRule rule_to_exclude: The one rule not to include in the copy
+        :param bool _ingress_rule: Whether the rule is an ingress or egress rule
+        :return: A copy of 'self' without the provided rule
+        :rtype: IstioSidecar
+        """
+        assert not _ingress_rule
+        res = IstioSidecar(self.name, self.namespace)
+        res.selected_peers = self.selected_peers
+        res.affects_egress = self.affects_egress
+        res.affects_ingress = self.affects_ingress
```
zivnevo, line 87:
```suggestion
        res = IstioSidecar(self.name, self.namespace)
        res.selected_peers = self.selected_peers
        res.affects_egress = self.affects_egress
        res.affects_ingress = self.affects_ingress
        res.default_sidecar = self.default_sidecar
```

network-config-analyzer/GenericYamlParser.py
```diff
@@ -186,3 +201,94 @@ def _get_connection_set_from_properties(dest_ports, method_set=MethodSet(True),
...
+            return None  # Not the relevant object
+        api_version = self.policy.get('apiVersion')
+        if 'istio' not in api_version:
+            return None  # apiVersion is not properly set
+        valid_keys = {'kind': [1, str], 'apiVersion': [1, str], 'metadata': [1, dict], 'spec': [0, dict]}
+        self.check_fields_validity(self.policy, policy_kind, valid_keys,
+                                   {'apiVersion': [istio_version]})
+        metadata = self.policy['metadata']
+        self.check_metadata_validity(metadata)
+        self.namespace = self.peer_container.get_namespace(metadata.get('namespace', 'default'))
```
zivnevo, line 246: User may get a false warning `Namespace istio-config is missing from the network configuration`

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        policy_name = self.parse_generic_istio_policy_fields('Sidecar', 'networking.istio.io/v1beta1')
+        if policy_name is None:
+            return None  # not relevant to build this policy
+
+        res_policy = IstioSidecar(policy_name, self.namespace)
+        res_policy.policy_kind = NetworkPolicy.PolicyType.IstioSidecar
+
+        if 'spec' not in self.policy or self.policy['spec'] is None:
+            self.warning('spec is missing or null in Sidecar ' + res_policy.full_name())
```
zivnevo, line 165: are we sure this is not a syntax error?
Better try on a live cluster

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,195 @@
...
+        sidecar_spec = self.policy['spec']
+        # currently, supported fields in spec are workloadSelector and egress
+        allowed_spec_keys = {'workloadSelector': [0, dict], 'ingress': [0, list], 'egress': [0, list],
+                             'outboundTrafficPolicy': [2, str]}
+        self.check_fields_validity(sidecar_spec, 'Sidecar spec', allowed_spec_keys)
+        res_policy.affects_egress = sidecar_spec.get('egress') is not None
+
+        workload_selector = sidecar_spec.get('workloadSelector')
+        if workload_selector is None:
+            res_policy.default_sidecar = True
```
zivnevo, line 177:
```suggestion
        res_policy.default_sidecar = workload_selector is None
```

zivnevo (changes requested):
thanks @shireenf-ibm. This is awesome.
please see some comments below

network-config-analyzer/GenericYamlParser.py
```diff
@@ -186,3 +201,94 @@ def _get_connection_set_from_properties(dest_ports, method_set=MethodSet(True),
         res = ConnectionSet()
         res.add_connections('TCP', tcp_properties)
         return res
+
+
+from PeerContainer import PeerContainer  # noqa: E402
```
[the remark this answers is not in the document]
shireenf-ibm: couldn't put it on the top, it creates a circular import (peer container uses the FilterActionType from the genericYamlParser)

network-config-analyzer/GenericYamlParser.py
```diff
@@ -186,3 +201,94 @@ def _get_connection_set_from_properties(dest_ports, method_set=MethodSet(True),
...
+
+
+class IstioGenericYamlParser(GenericYamlParser):
+    """
+    A class for istio yaml parser , common methods for istio policies parsers
+    """
+
+    def __init__(self, policy, peer_container, file_name=''):
```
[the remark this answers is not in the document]
shireenf-ibm: done

network-config-analyzer/GenericYamlParser.py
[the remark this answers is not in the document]
zivnevo: then better move `IstioGenericYamlParser` to a file of its own

network-config-analyzer/Peer.py
```diff
@@ -21,6 +21,8 @@ def __init__(self, name, namespace):
         self.namespace = namespace
         self.labels = {}  # Storing the endpoint's labels in a dict as key-value pairs
         self.extra_labels = {}  # for labels coming from 'labelsToApply' field in Profiles (Calico only)
+        self.ordered_specific_sidecars = []  # list of sidecars with workloadSelector selecting
```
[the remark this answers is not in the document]
shireenf-ibm: True

network-config-analyzer/IstioSidecarYamlParser.py
shireenf-ibm: true, it is a syntax error.
trying to apply on live cluster both authorization-policy and a sidecar without spec
got a server error for the sidecar wo the spec, while the auth-policy was created

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,195 @@
...
+        self._append_sidecar_into_relevant_list(str(res_policy), res_policy.selected_peers, res_policy.default_sidecar)
+
+        if sidecar_spec.get('ingress') is not None:
+            self.warning('Sidecar ingress is not supported yet.'
+                         f'Although "{res_policy.full_name()}" contains ingress entry, '
+                         f'it is not considered in policy connections', sidecar_spec)
+
+        egress = sidecar_spec.get('egress', [])
+        if egress is None:
+            self.syntax_error('An empty configuration is provided', res_policy.name)  # behavior of live cluster
```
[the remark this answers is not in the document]
shireenf-ibm: removed the syntax error, but still saving the specific sidecar as priority regardless it includes "egress" or not because:
While the spec states that, it seems to me that the live-cluster behaves differently.
This is what I've tested on live-cluster:
I've applied first a default sidecar on namespace : default
[def_sidecar_default_namespace1.txt](https://github.com/IBM/network-config-analyzer/files/9439292/def_sidecar_default_namespace1.txt)
and connections from this were really applied on all apps ` including ratings app`
Then I applied a new sidecar selecting the `ratings app` without egress specified, only ingress.
[sidecar_only_ingress.txt](https://github.com/IBM/network-config-analyzer/files/9439307/sidecar_only_ingress.txt)
but this specified sidecar "cancelled" the "egress" connections of the default sidecar. (no other sidecars or policies were applied).

image

I don't know whether my istio version is not updated. 
If we want to behave as written in the spec even though, I will call the `self._check_and_save_sidecar_if_top_priority(res_policy)` only if the sidecar contains egress rules. Should I do that?

network-config-analyzer/GenericYamlParser.py
shireenf-ibm: In last commit, I updated the get_namespace def to not warn on `istio_root_namespace` since it is a special case.
I also handled new cases regarding this namespace (referring to Noet2 and first example [here](https://istio.io/latest/docs/reference/config/networking/sidecar/)) as following:

- syntax error when global sidecar in root namespace includes workloadSelector.
(on live cluster, behavior of global sidecar with workloadSelector is undefined - as if the sidecar ignored)
- if a host's namespace is `'.'` and sidecar is in the root namespace , this means : allow egress traffic only to other workloads in the same namespace

zivnevo (changes requested):
I think we are almost there. Just these 2 things...

network-config-analyzer/IstioSidecarYamlParser.py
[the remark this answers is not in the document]
zivnevo: yes, the SideCar with no egress surely did something...  although I'm not sure I understand the output of `istioctl proxy-config clusters`. Did you also check the effect on connectivity? You may need to open a shell on the `ratings` container in the `ratings` Pod, and try for example to access the `reviews` service.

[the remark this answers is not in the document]
adisos: It seems the output of `istioctl proxy-config clusters` should be the list of services handled by the proxy for a pod. So it makes sense from this output that the rules from the 'egress sidecar' are not affecting the proxy of this pod after applying the 'ingress sidecar' (which does seem contradictory with spec definition). I do think it is worth checking actual connectivity as you suggested, and whether the sidecar ingress rule has some possible effect (maybe try as well an alternative sidecar with no ingress/egress rules instead of the one with ingress rule ).

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,187 @@
...
+        res_policy.affects_egress = sidecar_spec.get('egress') is not None
+
+        workload_selector = sidecar_spec.get('workloadSelector')
+        res_policy.default_sidecar = workload_selector is None
+        if str(self.namespace) == self.istio_root_namespace and workload_selector is not None:
+            self.syntax_error('Global Sidecar configuration should not have any workloadSelector.')
+        res_policy.selected_peers = self.update_policy_peers(workload_selector, 'labels')
+        self._check_and_save_sidecar_if_top_priority(res_policy)
+
+        for egress_rule in sidecar_spec.get('egress', []) or []:
```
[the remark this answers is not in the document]
shireenf-ibm: I kept this for "irregular" case of user's syntax error of  null egress field `egress:`
(example : `sidecar-invalid-egress.yaml `, `test: invalid_egress in sidecar_examples_w_onlineboutique/sidecars-syntax-tests-scheme.yaml` will be failed with  following error if we remove the `or []`:

>  for egress_rule in sidecar_spec.get('egress', []):
> TypeError: 'NoneType' object is not iterable

network-config-analyzer/Peer.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, name, namespace):
         self.namespace = namespace
         self.labels = {}  # Storing the endpoint's labels in a dict as key-value pairs
         self.extra_labels = {}  # for labels coming from 'labelsToApply' field in Profiles (Calico only)
+        self.prior_sidecar = None  # the first injected sidecar with workloadSelector selecting current peer
+        self.compare_namespaces_flag = False  # indicates if namespaces should be compared when checking global sidecar
+        # connections
```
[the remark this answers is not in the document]
shireenf-ibm: done

network-config-analyzer/IstioSidecarYamlParser.py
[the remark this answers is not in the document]
shireenf-ibm: - trying with another sidecar with a workloadSelector but without ingress and egress, got the same - the new sidecar overrides the whole default sidecar (the default's sidecar's egress is cancelled)

- testing connectivity on the shell did the same for me, the ratings pod could connect to services other than the ones defined in the default's sidecar egress 

It seems that a defined sidecar would override the default/global sidecar in any case (ignoring the existence of the 'egress' field ) - so leaving the code as is for now.

I have also published this contradicting "issue" on istio discussing forums. will update if I get a different answer/explanation

network-config-analyzer/IstioGenericYamlParser.py
```diff
@@ -0,0 +1,101 @@
...
+        if 'istio' not in api_version:
+            return None  # apiVersion is not properly set
+        valid_keys = {'kind': [1, str], 'apiVersion': [1, str], 'metadata': [1, dict], 'spec': [0, dict]}
+        self.check_fields_validity(self.policy, policy_kind, valid_keys,
+                                   {'apiVersion': [istio_version]})
+        metadata = self.policy['metadata']
+        self.check_metadata_validity(metadata)
+        ns_name = metadata.get('namespace', 'default')
+        warn_if_missing = ns_name != self.istio_root_namespace
+        self.namespace = self.peer_container.get_namespace(ns_name, warn_if_missing)
```
adisos, line 53: We may have some issue when changing the topology (peer container) based on sidecar with root namespace.
If some network config has such a sidecar resource, and we run forbids/permits query with a second network config with similar topology that does not have such sidecar resource, it may be not comparable with different topologies.

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,196 @@
...
+
+        return host_parts[0], host_parts[1]
+
+    def _get_peers_from_host_namespace(self, namespace):
+        """
+        returns all the services' peers in the specified namespace/s
+        The namespace can be set to *, ., or ~, representing any, the current, or no namespace, respectively
+        :param str namespace: the namespace value
+        :return: set of peers of services in the specified namespace value or empty PeerSet if no matching peers
+        :rtype: PeerSet
```
adisos, line 39: update docstring (method returns two values).

adisos (changes requested):
Added a few comments (did not yet finish to review all files).

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,196 @@
...
+        res_policy = IstioSidecar(policy_name, self.namespace)
+        res_policy.policy_kind = NetworkPolicy.PolicyType.IstioSidecar
+
+        if 'spec' not in self.policy or self.policy['spec'] is None:
+            self.syntax_error('spec is missing or null in Sidecar ' + res_policy.full_name())
+
+        sidecar_spec = self.policy['spec']
+        # currently, supported fields in spec are workloadSelector and egress
+        allowed_spec_keys = {'workloadSelector': [0, dict], 'ingress': [3, list], 'egress': [0, list],
+                             'outboundTrafficPolicy': [2, str]}
```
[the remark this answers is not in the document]
shireenf-ibm: Just for this version, since current version handles only internal services ...so this field is not supported yet. 
may change the code to 3 and skip using it for now ( in the next sidecar PR , this will be supported with adding support of external services)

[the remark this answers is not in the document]
adisos: ok, I think it is better to skip using code 3 for now.

network-config-analyzer/IstioGenericYamlParser.py
shireenf-ibm: would excluding the istio_root_namespace from the `__eq__` def in the `PeerContainer` help?
or maybe in order to make it general , I might save the namespaces which doesn't need to be warn about in `get_namespace`  into a `special_namespaces` dict and then in `__eq__` to exclude them from the "regular" namespaces comparison

adisos: Let's add a test to demonstrate this problem, and create a separate issue to resolve this later, in a different branch.
(Maybe we can handle conflicting topologies when only certain namespaces are missing, in relevant queries that currently do not support that).

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,196 @@
...
+                return
+            self.namespace.prior_default_sidecar = curr_sidecar
+            return
+
+        for peer in curr_sidecar.selected_peers:
+            if peer.prior_sidecar:  # this sidecar is not first one
+                self.warning(f'Peer "{peer.full_name()}" already has a Sidecar configuration selecting it. Sidecar: '
+                             f'"{curr_sidecar.full_name()}" will not be considered as connections for this workload')
+                return
+            peer.prior_sidecar = curr_sidecar
```
adisos, line 152: What cases does it cover of "another sidecar selecting this workload/namespace"? 
Shouldn't we issue error, for example, if two different sidecar objects select the same workload?
From istio documentation: 
"Each namespace can have only one Sidecar configuration without any workloadSelector that specifies the default for all pods in that namespace. It is recommended to use the name default for the namespace-wide sidecar. The behavior of the system is undefined if more than one selector-less Sidecar configurations exist in a given namespace. The behavior of the system is undefined if two or more Sidecar configurations with a workloadSelector select the same workload instance."

network-config-analyzer/IstioGenericYamlParser.py
shireenf-ibm: committed a new scheme file with a "not executed" query

adisos: great, thanks. 
please also create a new issue with some description of this problem + reference to this test example (which should be modified after the issue is resolved).

network-config-analyzer/IstioSidecarYamlParser.py
shireenf-ibm: On Live cluster, the "first" sidecar is considered always.  (they don't raise any errors / warnings)
So, we decided to raise a warning and consider the first sidecar (behave the same)

adisos: what is the "first" sidecar?

shireenf-ibm: I mean, for example, if you have two sidecars selecting the same workloads.
On live cluster, the sidecar that you apply first will be the only one considered for those workloads' connections. 
on our code, the first of them listed in our  `networkPolicyList` for a config will be considered ( a warning will be printed to tell that the other sidecar will not be considered) 
examples are in `bookinfo-sidecar-tests-scheme.yaml` :

1.  two-sidecars-with-same-workload-selector, query: connectivity-map-bookinfo-two-sidecars-with-same-workload-selector
2. (for default sidecars): 
        bookinfo-two-selector-less-sidecars, query : connectivity-map-bookinfo-ignoring-second-default-sidecar-in-same-namespace

network-config-analyzer/NetworkLayer.py
```diff
@@ -206,16 +206,15 @@ class IstioNetworkLayer(NetworkLayer):
     def _allowed_xgress_conns(self, from_peer, to_peer, is_ingress):
         # in istio applying default-allow if there is no capturing policy with action allow
         def captured_cond_func(policy):
-            return policy.action == IstioNetworkPolicy.ActionType.Allow
+            if isinstance(policy, IstioNetworkPolicy):
```
adisos, line 209:
```suggestion
            if policy.policy_kind == NetworkPolicy.PolicyType.IstioAuthorizationPolicy:
```

network-config-analyzer/NetworkLayer.py
```diff
@@ -206,16 +206,15 @@ class IstioNetworkLayer(NetworkLayer):
...
+                return policy.action == IstioNetworkPolicy.ActionType.Allow
+            return True  # sidecar always defines allowed connections
```
adisos, line 211:
```suggestion
            return True  # only for Istio AuthorizationPolicy the captured condition is more refined with 'Allow' policies
```

network-config-analyzer/IstioSidecar.py
```diff
@@ -0,0 +1,137 @@
...
+        conns = ConnectionSet(True)
+        # since sidecar rules include only peer sets for now, if a to_peer appears in any rule then connections allowed
+        for rule in self.egress_rules:
+            if to_peer in rule.egress_peer_set or \
+                    (to_peer in rule.special_egress_peer_set and from_peer.namespace == to_peer.namespace):
+                return PolicyConnections(True, allowed_conns=conns)
+
+        # egress from from_peer to to_peer is not allowed : if to_peer not been captured in the rules' egress_peer_set,
+        # or if the sidecar is global and to_peer is not in same namespace of from_peer while rule host's ns is '.'
+        return PolicyConnections(True, denied_conns=conns)
```
adisos, line 69: Shouldn't return here `PolicyConnections(True, allowed_conns=ConnectionSet())` ? (there is no 'deny' action for sidecar, so I think this should be considered as empty allowed connections rather than denied connections of All connections).

shireenf-ibm: True, fixing this

network-config-analyzer/IstioSidecarYamlParser.py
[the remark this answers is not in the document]
shireenf-ibm: Hi @zivnevo , @adisos ,
I got an answer in the following forum : [istio-discussion](https://discuss.istio.io/t/istio-sidecar-doc-description-vs-live-cluster-behavior/13849)

So from this answer I understand that the current code is fine (a specific sidecar always overrides a default one, regardless the existence of the 'egress' field)

adisos (commented):
About the tests: please also add tests with both sidecars and authorization policies in the network config. 
Also consider an example with resources from multiple layers (k8s, istio) as in the example from the link within the issue.

network-config-analyzer/IstioSidecarYamlParser.py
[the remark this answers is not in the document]
adisos: ok, please update the comment in the code with a link to this clarification.

adisos (commented):
Another question: is it possible to use some of the functionality of `IstioGenericYamlParser `within `IstioTrafficResourcesYamlParser.py` ?

shireenf-ibm:
> Another question: is it possible to use some of the functionality of `IstioGenericYamlParser `within `IstioTrafficResourcesYamlParser.py` ?

I don't see much in common
might call  `IstioGenericYamlParser.parse_generic_istio_policy_fields` from `IstioTrafficResourcesYamlParser.parse_gateway` to check the validity of the general fields (apiVersion, kind, metadata and spec)

shireenf-ibm:
> > Another question: is it possible to use some of the functionality of `IstioGenericYamlParser `within `IstioTrafficResourcesYamlParser.py` ?
> 
> I don't see much in common might call `IstioGenericYamlParser.parse_generic_istio_policy_fields` from `IstioTrafficResourcesYamlParser.parse_gateway` to check the validity of the general fields (apiVersion, kind, metadata and spec)

Actually I think this `parse_generic_istio_policy_fields` may be moved to `GenericYamlParser.py` and then all yamlParser objects we have may call it.

adisos:
> > > Another question: is it possible to use some of the functionality of `IstioGenericYamlParser `within `IstioTrafficResourcesYamlParser.py` ?
> > 
> > 
> > I don't see much in common. might call `IstioGenericYamlParser.parse_generic_istio_policy_fields` from `IstioTrafficResourcesYamlParser.parse_gateway` to check the validity of the general fields (apiVersion, kind, metadata and spec)
> 
> Actually I think this `parse_generic_istio_policy_fields` may be modified and moved to `GenericYamlParser.py` and then all `*yamlParser` objects we have may call it. @adisos , Should I do this here or in another, separate PR?

ok, good idea. 
you can do it on this branch.

tests/k8s_testcases/example_policies/services1/services1-scheme.yaml
```diff
@@ -5,25 +5,25 @@ networkConfigList:
   - name: ip
     networkPolicyList:
       - policies.yaml
-    expectedWarnings: 0
+    expectedWarnings: 1
```
adisos, line 8: why has the expected warning changed on this test?

network-config-analyzer/IstioTrafficResourcesYamlParser.py
```diff
@@ -51,17 +51,15 @@ def parse_gateway(self, gateway_resource, gateway_file_name):
...
-                                   {'apiVersion': [1, str], 'kind': [1, str],
-                                    'metadata': [1, dict], 'spec': [1, dict]})
-        if gateway_resource['kind'] != 'Gateway' or 'networking.istio.io' not in gateway_resource['apiVersion']:
-            return  # Not an Istio Gateway object
-        metadata = gateway_resource['metadata']
-        self.check_fields_validity(metadata, 'Gateway metadata', {'name': [1, str], 'namespace': [0, str]})
-        gtw_name = metadata['name']
-        gtw_namespace = self.peer_container.get_namespace(metadata.get('namespace', 'default'))
-        gtw_spec = gateway_resource['spec']
+        gtw_name, gtw_ns = self.parse_generic_yaml_objects_fields(gateway_resource, 'Gateway',
```
adisos, line 54: The second parameter for `parse_generic_yaml_objects_fields `should be of type `list`, but here it is called with `str`.

network-config-analyzer/GenericYamlParser.py
```diff
@@ -118,6 +132,38 @@ def check_fields_validity(self, dict_to_check, dict_name, allowed_keys, allowed_
             self.syntax_error(f'{bad_policy_keys.pop()} is not a valid entry in the specification of {dict_name}',
                               dict_to_check)
 
+    def parse_generic_yaml_objects_fields(self, yaml_object, object_kind, object_version, layer_keywords, spec_required=False):
+        """
+        Parse and check validity of the common fields in network yaml objects as policies and services.
+        e.g. kind, apiVersion, metadata and spec
+        :param dict yaml_object: the source dict of the network yaml object to be parsed
+        :param list object_kind : the possible kind value(s) of current object
```
adisos, line 140: When possible, be more specific with param type
```suggestion
        :param list[str] object_kind : the possible kind value(s) of current object
```

network-config-analyzer/K8sServiceYamlParser.py
```diff
@@ -24,20 +24,15 @@ def parse_service(self, srv_object):  # noqa: C901
...
-        if srv_object.get('kind') != 'Service' or srv_object.get('apiVersion') != 'v1':
-            return None  # Not a v1 Service object
-        metadata = srv_object.get('metadata')
-        if not metadata:
-            return None
-        srv_name = metadata.get('name')
-        if not srv_name:
-            return None
-        srv_namespace = metadata.get('namespace', 'default')
+        srv_name, srv_ns = self.parse_generic_yaml_objects_fields(srv_object, 'Service', ['v1'], ['v1', 'k8s'])
```
adisos, line 27: second parameter on function call should be a list and not str.

adisos (commented):
added a few more comments.

tests/k8s_testcases/example_policies/services1/services1-scheme.yaml
shireenf-ibm: because services in this test include `status` field .
I've added this field to the allowed keys with code 3 , so the new warning is :
`status is not yet supported inside Service,ignoring this key`

I found 'status' field in k8sIngress and Service objects in our tests, so I've added it to the ` allowed_keys` in case of 'k8s' type, otherwise, we'll get an error of `bad policy keys` (as an expected key to find)
 **maybe I should add this key with code 0 instead of 3?** 

However, I've added `status` as an allowed field for all `k8s` objects  since I found in k8s reference that this field is now supported in addition to `kind, spec, metadata and apiVersion` [(k8s_ref_example)](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/)

adisos: yes, it should be with code 0 instead of 3.
It was already supported in k8s netpols from a previous commit: 
https://github.com/IBM/network-config-analyzer/commit/9ff2b173561ab7958a7d13eb0db125b7e2e8a3f7

network-config-analyzer/GenericYamlParser.py
adisos: Other params on this function can be specified with more details on their type

network-config-analyzer/IstioGenericYamlParser.py
```diff
@@ -0,0 +1,76 @@
...
+
+        return res
+
+    def update_policy_peers(self, workload_selector, dict_key):
+        """
+        calculates the peers selected by current policy object
+        :param dict workload_selector: the workload selector of the policy object yaml
+        :param str dict_key: the key used in the selector for labels
+        :return the list of peers selected by the policy object
+        :rtype: list
```
adisos, line 66: returns a PeerSet and not a list

adisos (commented):
additional comment

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,195 @@
...
```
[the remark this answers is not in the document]
shireenf-ibm: true it should be `continue`

network-config-analyzer/IstioSidecarYamlParser.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Return the workload instances of the service in the given dns_name
+        :param str dns_name: the service given in the host
+        :rtype: PeerSet
+        """
+        # supported dns_name format is fqdn with internal k8s services pattern
+        # <service_name>.<domain>.svc.cluster.local
+        # the only relevant part of it is <service_name>
+        service_name = dns_name.split('.')[0]
+        return self.peer_container.get_pods_with_service_name_containing_given_string(service_name)
```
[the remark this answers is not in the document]
shireenf-ibm: fixing by adding a method that returns all pods with the service name

adisos (commented):
please also update `docs/IstioSupport.md ` with sidecar support, specify what is supported.

adisos (approved):

zivnevo (approved):
Big like for adding the `parse_generic_yaml_objects_fields()` method 🥇

PR #319, opened 2022-09-01
Issue 279 ns selector global np calico

shireenf-ibm:


network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -569,24 +569,25 @@ def _set_affects_ingress_egress(self, policy_spec, is_profile, res_policy):
         if not res_policy.affects_egress and 'egress' in policy_spec:
             self.syntax_error('A NetworkPolicy with egress field but no "Egress" in its policyTypes', policy_spec)
 
-    def _set_selected_peers(self, policy_spec, is_profile, res_policy):
+    def _set_selected_peers(self, policy_spec, is_profile, policy_name):
```
adisos, line 572: change function name to `_get_selected_peers` and also update docstring description (it does not "set" the selected peers member of the policy, but rather returns a set of the selected peers).

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -549,7 +549,7 @@ def _set_affects_ingress_egress(self, policy_spec, is_profile, res_policy):
         :return: None
         """
         allowed_policy_keys = {'order': 0, 'selector': 0, 'ingress': 0, 'egress': 0, 'types': 0, 'labelsToApply': 0,
-                               'doNotTrack': 3, 'preDNAT': 3, 'applyOnForward': 3}
+                               'namespaceSelector': 0, 'doNotTrack': 3, 'preDNAT': 3, 'applyOnForward': 3}
```
adisos, line 552: Maybe add an argument to this method to indicate if the policy is GlobalNetworkPolicy or not ,and only add the `namespaceSelector` to the `allowed_policy_keys` if it is global.

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -624,7 +623,14 @@ def parse_policy(self):
 
         policy_spec = self.policy['spec']
         self._set_affects_ingress_egress(policy_spec, is_profile, res_policy)
-        self._set_selected_peers(policy_spec, is_profile, res_policy)
+        res_policy.selected_peers = self._set_selected_peers(policy_spec, is_profile, res_policy.name)
+        ns_selector = policy_spec.get('namespaceSelector')
+        if ns_selector:
+            if not is_global_np:
+                self.syntax_error(f'namespaceSelector is not allowed in the spec of Calico {res_policy.policy_kind}')
```
adisos, line 630: if you update the function `_set_affects_ingress_egress ` as in my previous comment, we can remove this check here

adisos (commented):
Added a few comments.

Consider also adding a "bad path" test for a non-global network policy with a ns selector.

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -540,7 +540,7 @@ def _apply_extra_labels(self, policy_spec, is_profile, profile_name):
             for label, value in labels_to_apply.items():
                 pod.set_extra_label(label, value)
 
-    def _set_affects_ingress_egress(self, policy_spec, is_profile, res_policy):
+    def _set_affects_ingress_egress(self, policy_spec, is_profile, is_global_np, res_policy):
```
adisos, line 543: update docstring with new param

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -623,8 +624,11 @@ def parse_policy(self):
         res_policy.policy_kind = NetworkPolicy.get_policy_type_from_dict(self.policy)
 
         policy_spec = self.policy['spec']
-        self._set_affects_ingress_egress(policy_spec, is_profile, res_policy)
-        self._set_selected_peers(policy_spec, is_profile, res_policy)
+        self._set_affects_ingress_egress(policy_spec, is_profile, is_global_np, res_policy)
+        res_policy.selected_peers = self._get_selected_peers(policy_spec, is_profile, res_policy.name)
+        ns_selector = policy_spec.get('namespaceSelector')
+        if ns_selector:
+            res_policy.selected_peers &= self._parse_label_selector(ns_selector, policy_spec, namespace_selector=True)
```
adisos, line 631: Maybe this should be within `_get_selected_peers` : this method is supposed to return the peers selected by the policy, which should also take into account the `namespaceSelector`.

tests/calico_testcases/example_policies/testcase1/networkpolicy-with-ns-selector-bad-path-test.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: NetworkPolicy
+metadata:
+  name: bad-network-policy
+  namespace: kube-system
+spec:
+  selector: has(app)
+  namespaceSelector: unique-label == 'defaultNameSpace'
+  types:
+  - Ingress
+  - Egress
```
adisos, line 11: Shouldn't this policy be in a scheme file `networkConfig ` with `expectedError`?

adisos (changes requested):
Added a few more comments.

tests/calico_testcases/example_policies/testcase1/networkpolicy-with-ns-selector-bad-path-test.yaml
shireenf-ibm: I've added it as the last `networkConfig ` in `tests/calico_testcases/example_policies/testcase1/testcase1-scheme.yaml`

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -580,17 +581,23 @@ def _get_selected_peers(self, policy_spec, is_profile, policy_name):
...
             if is_profile:
                 return self.peer_container.get_profile_pods(policy_name, True)
             else:
                 return self.peer_container.get_namespace_pods(self.namespace)
 
+        ns_selector = policy_spec.get('namespaceSelector')
+        if ns_selector:
+            peers &= self._parse_label_selector(ns_selector, policy_spec, namespace_selector=True)
+        return peers
+
```
adisos, line 600: Is it possible to have a global network policy with a namespace selector and  without a selector ? if yes, it seems this current version may miss this.

shireenf-ibm: I couldn't understand if this is possible from their docs.
from the "old" code, I concluded that before the namespace selector , a GlobalNetworkpolicy had to contain a selector. (all the cases under `else` doesn't match for it being a global networkPolicy (only profile or networkpolicy with a namespace)  and assumed the selector is still a must field.
I will try to test that on live cluster and fix if needed

adisos: ok.
About the "cases under `else `doesn't match for it being a global networkPolicy" - I think it does match, because `self.namespace` can be `None ` for a global np.

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -569,24 +572,32 @@ def _set_affects_ingress_egress(self, policy_spec, is_profile, res_policy):
...
-            res_policy.selected_peers = self._parse_label_selector(pod_selector, policy_spec, self.namespace)
+            peers &= self._parse_label_selector(pod_selector, policy_spec, self.namespace)
         else:
             if is_profile:
-                res_policy.selected_peers = self.peer_container.get_profile_pods(res_policy.name, True)
-            else:
-                res_policy.selected_peers = self.peer_container.get_namespace_pods(self.namespace)
+                return self.peer_container.get_profile_pods(policy_name, True)
+            elif not is_global_np:
+                return self.peer_container.get_namespace_pods(self.namespace)
```
adisos, line 595:
```suggestion
            else:
                peers = self.peer_container.get_namespace_pods(self.namespace)
```
I think we can apply this change and avoid having to check if this is global network policy.

network-config-analyzer/CalicoPolicyYamlParser.py
```diff
@@ -569,24 +572,32 @@ def _set_affects_ingress_egress(self, policy_spec, is_profile, res_policy):
...
+
+        ns_selector = policy_spec.get('namespaceSelector')
```
adisos, line 597:
```suggestion
        # consider namespaceSelector for GlobalNetworkPolicy when it exists
        ns_selector = policy_spec.get('namespaceSelector')
```

adisos (approved):

PR #328, opened 2022-09-08
handling the case of err output of a live cluster cmd

shireenf-ibm:
explanation of this commit:

An executable command (kubectl, calicoctl or helm) called from nca runs as a subprocess.
A subprocess may **success** and return a stdout - which is redirected to a Pipe for handling later (reading the yaml resources)
or it may **fail** returning an _empty stdout_ and err messages to the _stderr_ which **was** always redirected to the sys.stderr.
In this commit I redirect the errors of an executable command to a pipe too.(since we don't want to print them always)

 we have 2 cases for calling the executables:
 1. we are sure 100% that the user intended to run the live cluster to get the resources (by passing k8s / no inputs/ calico / istio / helm), for these cases , if an error was returned by the command-line , it will be printed
 2. the user provided paths by using the (`-r , resourceList flag`) which don't include all wanted resources - so we try to fill the containers from the live-cluster. Since a live-cluster may not exist/ disconnected - our attempt should be done silently , and errors of running the commands above will be ignored (running as if 0 resources were found)


shireenf-ibm:
In first commit implemented "trying to load from live cluster" silently
after seing task 2 in the issue , I commented a new addition : _if the resource list refers to a single dir that contains no policies don't try to load them from live cluster_
In my opinion, it is worth thinking if we have any case that we should try to load policies from live cluster if `resourceList ` paths don't include any!!

shireenf-ibm:
reverted changes from previous commits .
Cancelled the option of trying to load from live cluster in case resources are missing from `resource_list` paths (using `-r` means stick to the given paths)
+ updated relevant doc
---------------------------------------------------------
loading from live cluster is now possible when:
1. layer name is provided as a path by the user 
2. the user doesn't use any flag (example : `--connectivity`) 
3. _The old behavior_ with specific flags only (no use of `-r`). example:  when a user provides only policy list (i.e.: `--connectivity <policies_path>) then topology will be loaded from k8s

nca/NetworkConfig/ResourcesHandler.py
```diff
@@ -81,8 +82,13 @@ def _set_config_peer_container(self, ns_list, pod_list, resource_list, config_na
             # deepcopy is required since PoliciesFinder may change peer_container
             peer_container = copy.deepcopy(self.global_peer_container)
         else:  # the specific networkConfig has no topology input resources (not private, neither global)
-            print('loading topology objects from k8s live cluster')
-            resources_parser.load_resources_from_k8s_live_cluster([ResourceType.Namespaces, ResourceType.Pods])
+            # this case is reachable when:
+            # 1. no input paths are provided at all, i.e. user intended to get resources from live cluster
+            # 2. paths are provided only using resourceList flag, but no topology objects found;
```
adisos, line 87: "paths are provided only using resourceList flag" - why do we load from live cluster on this case?

adisos: ok, noticed we do not load on this case

adisos (approved):

PR #332, opened 2022-09-15
comparable peer containers

shireenf-ibm:
configurations may be considered identical when having same set of peers and policies

Signed-off-by: shireenf <shireenf@il.ibm.com>


nca/NetworkConfig/PeerContainer.py
```diff
@@ -37,6 +37,14 @@ def __eq__(self, other):
...
         return False
 
+    def is_comparable_with_other_container(self, other):
+        """
+        Two peer containers are considered comparable if having same set of peers, regardless their namespaces
+         and services
+         :param PeerContainer other: the peer container to compare current with
+        """
+        return isinstance(other, PeerContainer) and self.peer_set == other.peer_set
+
```
adisos, line 47: I think we should consider the peer containers as comparable only if both peer_set and services are identical (thus it is comparable regardless of namespaces only).

For example, if an Ingress resource uses some service name, but the service is different between the two configs (captures different pods for example), then this common Ingress policy has different connectivity results on each config, and this cannot fall into the category of "have the same network topology and the same set of policies"

nca/NetworkConfig/PeerContainer.py
```diff
@@ -37,6 +37,14 @@ def __eq__(self, other):
                 and self.services == other.services
         return False
 
+    def is_comparable_with_other_container(self, other):
+        """
+        Two peer containers are considered comparable if having same set of peers, regardless their namespaces
+         and services
```
adisos, line 43: also update the documentation

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -714,8 +714,8 @@ def __init__(self, config1, config2, output_config_obj=None):
     def get_query_type():
         return QueryType.ComparisonToBaseConfigQuery
 
-    def is_identical_topologies(self, check_same_policies=False):
-        if self.config1.peer_container != self.config2.peer_container:
+    def is_identical_configs(self, check_same_policies=False):
```
adisos, line 717: I would keep the old name `is_identical_topologies`, as the term 'config' also refers to policies

shireenf-ibm: Ok, although I changed the name because in its second condition we check for eq. policies (although we get there only with the true flag)

adisos (approved):

PR #334, opened 2022-09-20
Added complex ingress tests.

tanyaveksler:
Signed-off-by: Tanya <tatyana@il.ibm.com>


adisos (changes requested):
Hi @tanyaveksler , looks good.
I think we should also have an additional scheme file for k8s ingress test which takes both of the ingress resource yamls in the policy list.

adisos (approved):

PR #341, opened 2022-10-02
print all counter examples 

shireenf-ibm:


nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -767,17 +767,22 @@ def exec(self, layer_name=None):
...
-                    return QueryAnswer(False, self.name1 + ' and ' + self.name2 + ' are not semantically equivalent.',
-                                       explanation)
+                    different_conns_list.append(f'Allowed connections from {peer1} to {peer2}'
+                                                f' are different\n' + conns1.print_diff(conns2, self.name1, self.name2))
+        if different_conns_list:
+            explanation = different_conns_list[0] + '\n'
+            if self.output_config.printAllPairs:
+                explanation += '\n'.join(different_conns_list[1:])
+            return QueryAnswer(False, self.name1 + ' and ' + self.name2 + ' are not semantically equivalent.\n',
+                               explanation)
```
adisos, line 785: I think that for efficiency considerations, it may be better to break from the loop if the output config only requires one counterexample. (relevant for all queries where this applies to).

adisos (changes requested):
Hi @shireenf-ibm , looks good. Added some comment on the code. 
Also, another comment about the output:

Example current output:
```
Running query twoWayContainment_a_with_different_ipBlock_policies_print_all_pairs
Neither network configuration config_a_with_ipBlock and config_a_with_b_ipBlock are contained in the other.
config_a_with_ipBlock is not contained in config_a_with_b_ipBlock:

Allowed connections from 10.12.0.0-10.255.255.255 to default/pod-2 in config_a_with_ipBlock are not a subset of those in config_a_with_b_ipBlock
config_a_with_ipBlock allows all connections while config_a_with_b_ipBlock does not.
Allowed connections from default/pod-4 to default/pod-2 in config_a_with_ipBlock are not a subset of those in config_a_with_b_ipBlock
config_a_with_ipBlock allows all connections while config_a_with_b_ipBlock does not.
Allowed connections from default/pod-0 to default/pod-2 in config_a_with_ipBlock are not a subset of those in config_a_with_b_ipBlock
config_a_with_ipBlock allows all connections while config_a_with_b_ipBlock does not.
Allowed connections from 10.10.0.0-10.10.255.255 to default/pod-2 in config_a_with_ipBlock are not a subset of those in config_a_with_b_ipBlock
config_a_with_ipBlock allows all connections while config_a_with_b_ipBlock does not.
...
```

I think a more clear output would be as follows:
```
config_a_with_ipBlock is not contained in config_a_with_b_ipBlock:
Connections allowed in config_a_with_ipBlock which are not a subset of those in config_a_with_b_ipBlock:
src: 10.12.0.0-10.255.255.255, dst: default/pod-2, conn: All connections
src: default/pod-4, dst: default/pod-2, conn: All connections
...
```

docs/SchemeFileFormat.md
```diff
@@ -78,12 +78,13 @@ For example: `my_set/prod_ns/deny_all_policy`. If there are multiple policies na
...
-|subset| A dict object with the defined subset elements to display in the output|[subset](#subset) object|
+| Field           | Description                                                                      | Value                                  |
+|-----------------|----------------------------------------------------------------------------------|----------------------------------------|
+| outputFormat    | Output format specification.                                                     | string [ txt / yaml / csv / md / dot ] |
+| outputPath      | A file path to redirect output into.                                             | string                                 |
+| outputEndpoints | Choose endpoints type in output.                                                 | string [ pods / deployments ]          |
+| subset          | A dict object with the defined subset elements to display in the output          | [subset](#subset) object               |
+| printAllPairs   | Choose if to print all counter peer pairs causing the query result in the output | bool                                   |
```
zivnevo, line 87: I suggest to replace `PrintAllPairs` with something like `FullExplanation`. This will be handy for (future) queries where the explanation might look different.
Note that the explanation is usually more than just a pair of peers. It also shows protocols, etc.

zivnevo (changes requested):
Looks good.
Please also add expectedOutput to one of the tests. I believe this should work for queries other than connectivity and diff.
It may require to sort the results though.

docs/SchemeFileFormat.md
```diff
@@ -78,13 +78,13 @@ For example: `my_set/prod_ns/deny_all_policy`. If there are multiple policies na
...
-| printAllPairs   | Choose if to print all counter peer pairs causing the query result in the output | bool                                   |
+| Field           | Description                                                                    | Value                                  |
+|-----------------|--------------------------------------------------------------------------------|----------------------------------------|
+| outputFormat    | Output format specification.                                                   | string [ txt / yaml / csv / md / dot ] |
+| outputPath      | A file path to redirect output into.                                           | string                                 |
+| outputEndpoints | Choose endpoints type in output.                                               | string [ pods / deployments ]          |
+| subset          | A dict object with the defined subset elements to display in the output        | [subset](#subset) object               |
+| FullExplanation | Choose if to print all counter examples causing the query result in the output | bool                                   |
```
adisos, line 87:
```suggestion
| FullExplanation | Choose if to print all counterexamples causing the query result in the output | bool                                   |
```

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -777,10 +777,10 @@ def exec(self, layer_name=None):
                 if conns1 != conns2:
                     different_conns_list.append(f'Allowed connections from {peer1} to {peer2}'
                                                 f' are different\n' + conns1.print_diff(conns2, self.name1, self.name2))
+                    if not self.output_config.FullExplanation:
+                        break
```
adisos, line 781: Note this is a nested loop, and in python the `break` keyword can only break out of the inner-most loop. But in this case we need to break from both loops.

zivnevo (changes requested):
@shireenf-ibm , this is great. I really like the `_query_answer_with_relevant_explanation` functions. They will help us in the future 👍 

Please change `FullExplanation` to `fullExplanation` so the capitalization is consistent with the other output settings (sorry, I should have noticed this earlier).

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -101,7 +101,8 @@ def run_query(self, cmd_line_flag=False):
         comparing_err = 0
         self.output_configuration.print_query_output(query_output, formats)
         if self.expected_output_file is not None:
-            if self.query_name in {'ConnectivityMapQuery', 'SemanticDiffQuery'}:
+            if self.query_name in {'ConnectivityMapQuery', 'SemanticDiffQuery'} or\
+                    self.output_configuration.FullExplanation:
```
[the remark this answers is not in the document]
shireenf-ibm: I hesitated to generalize this, since when `FullExplanation` is False we can not predict what is the one example that will be printed (so I wanted to avoid using expectedOutput when full explanation is not used).

Also, to avoid using it with queries that were not touched in this PR (queries derived from `NetworkConfigQuery`) - maybe their outputs needs to be sorted too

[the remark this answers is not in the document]
zivnevo: OK, got it. Leave it as-is then.

zivnevo (approved):

PR #349, opened 2022-10-18
support yaml outputs for  more queries 

shireenf-ibm:
Hi @adisos , @zivnevo ,
This PR is still in progress (may be not ready for code review) but I want to show you the new 'yaml' output format of some queries and get your comments regarding it

Notes:
- In this commit changes (supporting yaml output) are applied only to queries derived from` TwoNetworkConfigsQuery` (still did not apply any changes for queries derived from `NetworkConfigQuery`)
- currently, I didn't make any changes to the `yaml` output format of `SemanticDiffQuery` since it  was supported previously
- The main idea of making all other ` TwoNetworkConfigsQuery` support yaml format in addition to txt format (and with possibility to support other output formats) is by "returning" the query's output explanation as list of "tuples" instead of "txt" (str). 
and converting it into the required format later when computing the query output
- For your review, this commit  includes examples of queries running with `outputFormat: yaml` and their expected output files 


adisos:
> Hi @adisos , @zivnevo , This PR is still in progress (may be not ready for code review) but I want to show you the new 'yaml' output format of some queries and get your comments regarding it
> 
> Notes:
> 
>     * In this commit changes (supporting yaml output) are applied only to queries derived from` TwoNetworkConfigsQuery` (still did not apply any changes for queries derived from `NetworkConfigQuery`)
> 
>     * currently, I didn't make any changes to the `yaml` output format of `SemanticDiffQuery` since it  was supported previously
> 
>     * The main idea of making all other ` TwoNetworkConfigsQuery` support yaml format in addition to txt format (and with possibility to support other output formats) is by "returning" the query's output explanation as list of "tuples" instead of "txt" (str).
>       and converting it into the required format later when computing the query output
> 
>     * For your review, this commit  includes examples of queries running with `outputFormat: yaml` and their expected output files
> 
> 
> Signed-off-by: shireenf-ibm [shireenf@il.ibm.com](mailto:shireenf@il.ibm.com)

Hi @shireenf-ibm , the yaml output format looks very good to me.

zivnevo:
@shireenf-ibm , I think the `explanation` field should be of type `dict` with two possible keys: `description` and `connections`. Currently its type is list. Otherwise, looks good.

shireenf-ibm:
more updates
- expanded support of full explanation to more queries (queries derived from `NetworkConfigQuery`)
- now : all queries support `yaml` output format except `SanityQuery` (sanity query may have many different explanations - prefered to keep it in txt format only)
- added examples for all queries supporting yaml (+ their expectedOutput files)
- didn't change the `yaml outputFormat` of `SemanticDiffQuery `and `ConnectivityMapQuery`

shireenf-ibm:
last commit includes `json` support with expected output examples
json reults are stored in a top list ([]), so if a specific query is called several times (on multiple configs/pairs) all results are included in one top json object (otherwise json file is not legal) - example: `tests/calico_testcases/expected_output/equiv-all-range1.json`

adisos (changes requested):
hi @shireenf-ibm , a few comments:
- separate json format to another branch/PR (does it suppoer conectivity map/ semantic diff?) 
- execute_and_write_output_in_required_format : 
	- consider common function to be used
	- document function: input params, rtype, ret values 
- write_query_output: missing documentation of rtype, returned values 
- confusing names: write_query_output() vs get_query_output()  ('write' - misleading name, output is not written, just returned as string)

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1360,15 +1705,20 @@ def exec(self):
...
+            query_answer.output_result += f'\n{self.output_config.secondConfigName} forbids connections specified in ' \
+                                          f'{self.output_config.configName}'
+        if query_answer.output_explanation:
+            query_answer.output_result = f'{self.output_config.secondConfigName} does not forbid connections specified ' \
+                                         f'in {self.output_config.configName}'
+        explanation_header = f'Both {self.output_config.configName} and ' \
+                             f'{self.output_config.secondConfigName} allow the following connection(s)'
+        return query_result,\
+            self.write_query_output(query_answer, explanation_header, [explanation_header], int(query_result)), \
+            query_answer.query_not_executed
```
adisos, line 1721: Can try to simplify this as follows? 
extend `exec `function to consider the cmd line flag:

```
    def exec(self, cmd_line_flag=False):
        if not self.config1:
            return QueryAnswer(False, 'There are no NetworkPolicies in the given forbids config. '
                                      'No traffic is specified as forbidden.', query_not_executed=True)
        if self.config1.policies_container.layers.does_contain_single_layer(NetworkLayerName.Ingress):
            return QueryAnswer(bool_result=False,
                               output_result='Forbidden traffic cannot be specified using Ingress resources only',
                               query_not_executed=True)

        config1_without_ingress = self.clone_without_ingress(self.config1)

        res = IntersectsQuery(config1_without_ingress, self.config2, self.output_config).exec(True)
        res.bool_result = not res.numerical_result if cmd_line_flag else res.bool_result
        if res.numerical_result == 1:
            res.output_result += f'\n{self.output_config.secondConfigName} forbids connections specified in ' \
                                          f'{self.output_config.configName}'
        if res.output_explanation:
            res.output_result = f'{self.output_config.secondConfigName} does not forbid connections specified ' \
                                         f'in {self.output_config.configName}'
        return res
```

and simplify `execute_and_write_output_in_required_format` as follows:

```
    def execute_and_write_output_in_required_format(self, cmd_line_flag=False):
        query_answer = self.exec(cmd_line_flag)
        explanation_header = f'Both {self.output_config.configName} and ' \
                             f'{self.output_config.secondConfigName} allow the following connection(s)'
        return query_answer.bool_result,\
            self.write_query_output(query_answer, explanation_header, [explanation_header], int(query_answer.bool_result)), \
            query_answer.query_not_executed
```

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1131,27 +1413,47 @@ def exec(self, only_captured=False):
...
+        the query_answer.output_explanation list
+        The output explanation of the ContainmentQuery may be :
+        (1) list[str] - when configs containment is not fulfilled because of different peers
+        (2) list[tuple] - when configs containment is not fulfilled because of different connections between same peers
+        :param str config_name_1: the name of the first config
+        :param str config_name_2: the name of the second config
+        :param first_elem: the first element of the output_explanation to determine its type
+        :rtype: str
+        """
+        if isinstance(first_elem, str):
```
adisos, line 1453: I would consider changing the numerical result of this query to determine if it is not contained due to pods or actual connections. (such as 0 for not contained due to connections, 1 for containment, 2 for not contained due to pods).
So we don't have to rely on the type of the elements in the output explanation to determine this.

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1166,37 +1468,83 @@ def get_query_type():
...
             return QueryAnswer(bool_result=True,
                                output_result=f'The two network configurations {self.name1} and {self.name2} '
                                              'are semantically equivalent.',
-                               numerical_result=3)
+                               numerical_result=3), None, None
+
+        sub_query_answer1 = QueryAnswer(output_result=contained_1_in_2.output_result,
+                                        output_explanation=contained_1_in_2.output_explanation)
+        sub_query_answer2 = QueryAnswer(output_result=contained_2_in_1.output_result,
+                                        output_explanation=contained_2_in_1.output_explanation)
```
adisos, line 1484: Why do you have to recreate `QueryAnswer `objects here and cannot use the objects returned from running `exec `on `ContainmentQuery`?

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1166,37 +1468,83 @@ def get_query_type():
...
         # (contained_2_in_1)
         return QueryAnswer(bool_result=False,
-                           output_result=f'Network configuration {self.name2} is a proper subset of {self.name1}.\n',
-                           output_explanation=explanation_not_contained_self_other, numerical_result=1)
+                           output_result=f'Network configuration {self.name2} is a proper subset of {self.name1}',
+                           numerical_result=1), sub_query_answer1, None
+
+    def execute_and_write_output_in_required_format(self, cmd_line_flag=False):
+        # TwoWayContainmentQuery executes two Containment sub queries
+        query_answer, sub_query_answer_1, sub_query_answer_2 = self.exec()
```
adisos, line 1501: cannot you use the numerical result from `query_answer `instead of having to use `sub_query_answer_1 `and `sub_query_answer_2 `?

adisos (commented):
Hi @shireenf-ibm , added a few more comments.

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -61,6 +64,49 @@ def policy_title(policy):
...
+        calls the relevant method to write the query output in the required format
+        :param QueryAnswer query_answer: the query answer
+        :param str explanation_header: the txt message that describes the query's output explanation
+        :param list yaml_explanation_descriptions: [optional] the txt messages of explanations for yaml format,
+        relevant for TwoNetworkConfigsQuery classes
+        :param int query_result: [optional] the numerical result of the query, relevant for TwoNetworkConfigsQuery classes
+        """
+        # compatibility for the called methods
+        yaml_description = yaml_explanation_descriptions or explanation_header
+        numerical_result = query_result if query_result is not None else int(query_answer.numerical_result)
```
adisos, line 79: Why not having the correct/final `numerical_result `within `query_answer`?
One option could be to re-assign `query_answer.numerical_result` with `query_result` if exists.
A possibly better option would be that all queries return `query_answer `that already contains the correct `numerical_result`.
In any case , `write_yaml_output `does not need this extra argument if it can be used from  `query_answer`.

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -61,6 +64,49 @@ def policy_title(policy):
...
+    def write_txt_output(self, query_answer, explanation_description):
+        query_output = query_answer.output_result + '\n'
+        if query_answer.output_explanation:
+            query_output += self.compute_txt_explanation(query_answer, explanation_description) + '\n'
+        return query_output
+
+    def write_yaml_output(self, query_answer, descriptions, query_result):
+        query_name = self.output_config.queryName or type(self).__name__
+        configs_field = 'config' if isinstance(self, NetworkConfigQuery) else 'configs'
+        yaml_content = {'query': query_name, configs_field: self.get_configs_names()}
```
adisos, line 97: I think that we can just use `configs `with a list of config names , whether it is 1 config or 2 configs.
`get_configs_names` would then return a list in both cases.
It will also simplify this code to one line:
`yaml_content = {'query': query_name, 'configs': self.get_configs_names()}`

nca/NetworkConfig/NetworkConfigQuery.py
shireenf-ibm: actually I can use them, I just need the `output_result` and `output_explanation` of them for 'computing' the `query_output`. I thought that way is easiest to read and understand the "purpose" of them.
I will simplify this (by not creating new objects)

nca/NetworkConfig/NetworkConfigQuery.py
shireenf-ibm: as mentioned above, here I use the `sub_query_answer_1` and `sub_query_answer_2` as objects for storing the `output_result` and `output_explanation` data of each `ContainmentQuery.exec()` Call.
as I need them for writing the correct output 
i.e. if both config1 is not contained in config2 and config2 is not contained in config1 , I need the separate explanations of each ContainmentQuery.
example is in : `tests/k8s_testcases/expected_output/two-way-containment-a-with-different-ipblock-policies-all-pairs.yaml`
The first part of the explanation is the `sub_query_answer_1.output_explanation` which is the result of the first call of `ContainmentQuery.exec()`
and the second part is the  `sub_query_answer_2.output_explanation` which is the result of the second call of `ContainmentQuery.exec()`

nca/NetworkConfig/NetworkConfigQuery.py
shireenf-ibm: agree, this actually  better . I didn't do that in order not to change the old `exec`. 
but it seems that changes on "building" the `QueryAnswer ` fields in `exec()` methods may help a lot to get a cleaner code

nca/NetworkConfig/NetworkConfigQuery.py
adisos: ok, but wouldn't it be more clear to identify each case based on the `numerical_result `of `query_answer` (between 0 to 3) instead of comparing sub_result to `None`?

shireenf-ibm: True, I'll simplify this too

nca/Utils/QueryOutputHandler.py
```diff
@@ -0,0 +1,360 @@
...
+
+@dataclass
+class QueryAnswer:
+    """
+    A class for holding the answer to any one of the NetworkConfigQuery.py queries
+    """
+    bool_result: bool = False
+    output_result: str = ''
+    explanation_description: str = ''
+    output_explanation: OutputExplanation = None
```
adisos, line 19: consider adding `explanation_description `to `OutputExplanation`.

adisos (changes requested):
Hi @shireenf-ibm , good progress. Added a few more comments.

nca/Utils/QueryOutputHandler.py
```diff
@@ -0,0 +1,358 @@
...
+            self.write_policies_and_rules_explanations(explanation.explanation_description,
+                                                       explanation.policies_and_rules)
+        elif explanation.pods_lists:
+            self.write_pods_list_explanation(explanation.explanation_description, explanation.pods_lists)
+        elif explanation.connections_diff:
+            self.write_conns_diff_explanation(explanation.explanation_description, explanation.connections_diff)
+        elif explanation.combined_explanation:
+            self.handle_explanation_by_type(explanation.combined_explanation.two_results_combined[0])
+            self.handle_explanation_by_type(explanation.combined_explanation.two_results_combined[1])
+
```
adisos, line 105: Can we avoid having a separate method for handling each type of explanation? 
Maybe the `OutputExplanation ` object itself can have a method that returns its `txt `or `yaml ` str, and then you can update the relevant member of the OutputHandler with the result.

adisos: Is it possible to compute the `dict` object from the explanation separately, so then it can be used for creating `yaml `or `json `output?

adisos (changes requested):

nca/Utils/QueryOutputHandler.py
shireenf-ibm: > Is it possible to compute the `dict` object from the explanation separately, so then it can be used for creating `yaml `or `json `output?

yes, my plan when adding "json" support is that JsonOutputHandler , may be derived from the YamlOutputHandler and the to separate the "compute" and "dump" parts so only the dump will be different

adisos: good, so let's do the separation already on this branch.

shireenf-ibm: > Can we avoid having a separate method for handling each type of explanation? Maybe the `OutputExplanation ` object itself can have a method that returns its `txt `or `yaml ` str, and then you can update the relevant member of the OutputHandler with the result.

hi @adisos , for your review, in last commit , each explanation's type dataclass is responsible for computing the yaml and txt outputs (had to write 2 def for each) 
syaml explanation is returned as real "yaml" field (list[dict]) to be added to the other yaml fields and then all together dumped into yaml string

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -775,28 +828,24 @@ def exec(self, layer_name=None):
                 conns1, _, _, _ = self.config1.allowed_connections(peer1, peer2, layer_name)
                 conns2, _, _, _ = self.config2.allowed_connections(peer1, peer2, layer_name)
                 if conns1 != conns2:
-                    different_conns_list.append(f'src: {peer1}, dst: {peer2}, description: '
-                                                f'{conns1.print_diff(conns2, self.name1, self.name2)}')
+                    different_conns_list.append((str(peer1), str(peer2), conns1, conns2))
```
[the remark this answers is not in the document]
shireenf-ibm: Hi @adisos , 
I got a little confused with this.
conns1 is the allowed connections from src to dst in config1
conns2 is the allowed connections from same src to dst in config2

On branch master (which supported only txt output) : 
it print the diff by  `conns1.print_diff(conns2, self.name1, self.name2`, **I do the same in `TxtOutputHandler.write_conns_diff_explanation`** - i.e. calling same `print_diff` (from connectionSet, on same parameters)

However, On YamlOutputHandler , I write **two** yaml objects for the same query results,
with the output result and explanation telling that the connections are different between the 2 configs
and then provide once the connections from the first config (conns1) 
and the other yaml object with conns2 
So do you suggest writing the yaml output in one result , by writing a new diff method? (for both txt and yaml results)

tests/k8s_testcases/expected_output/no-strong-equivalence-all-peers-pairs.yaml
```diff
@@ -0,0 +1,336 @@
...
+    conn: All connections
+  - src: kube-system/calico-node-mgdlr
+    dst: kube-system/file-plugin-7bfb8b69bf-p86gk
+    conn: All connections
+  - src: kube-system/calico-node-mgdlr
+    dst: kube-system/heapster-7df8cb8c66-zxkk2
+    conn: All connections
+  - src: kube-system/calico-node-mgdlr
+    dst: kube-system/keepalived-watcher-57ghx
+    conn: All connections
```
adisos, line 26: I would avoid having 2 yaml objects for the equivalence query result.
We can have one object, similar to the result of the 2-way-containment query.

nca/NetworkConfig/NetworkConfigQuery.py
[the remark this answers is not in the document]
adisos: I notice some inconsistency for the output content in `txt` vs `yaml` formats on this `Equivalence` query:
- For `txt` with `fullExplanation` flag, the output contains a list of `(src,dst)` pairs, where for each pair there is an **example** for a connection allowed in one config but not in the other. (formatted by `print_diff` of `ConnectionSet`).
- For `yaml` with `fullExplanation` flag, the output contains 2 lists : one with `(src,dst)` attached to all allowed connections from `src `to `dst `in `config1`, and the other list is similar, only with all connections allowed in `config2`.

tests/k8s_testcases/expected_output/no-strong-equivalence-all-peers-pairs.yaml
shireenf-ibm: okay, so I'll resolve this first and after that handle the "_diff_" issue

nca/NetworkConfig/NetworkConfigQuery.py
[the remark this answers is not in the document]
zivnevo: If it's only about `Equivalence` query, I don't really care how the output should look like. We should seriously consider if this query should be exposed to users, as both `SemanticDiffQuery` and `TwoWayContainmentQuery` are far more useful.
Note that the query is still being used as a macro for other queries, so we can't dispose it, but I don't think the other queries will use its output.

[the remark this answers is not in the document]
zivnevo: In `InterferesQuery` we only care about allowed connections in config1 which are not allowed by config2 (and which config2 allows them explicitly). So I think one list is enough.
btw, the variable names in this query are a mess: 
```python
_, captured1_flag, conns1_captured, _ = self.config2.allowed_connections(peer1, peer2)
```
is really confusing (variables have index 1 but refer to `config2`).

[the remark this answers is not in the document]
shireenf-ibm: Trying to write a "compute_diff" def for connectionSet, I realized this:
- when all connections are allowed, in `ConnectionSet `, there is an `allow_all` flag and the connections map is cleared (so I assume, it is preferred not to print the list of all possible connections)
- when an extended connection is AllConnections but the "other" connection includes only one protocol/property, the diff may be a long (maybe ambiguous) list

In `print_diff` (for textual result) we simply print:

`<config_1> allows all connections while <config_2> does not.`

so a new idea (examples were committed last) to represent the conns_diff between two configs for `InterferesQuery` and `EquivalenceQuery` is writing both connections in same explanation list, i.e.

- src: <src_peer_name>
  dst: <dst_peer_name>
  conns in <config_1>: <conns>
  conns in <config_2>: <conns>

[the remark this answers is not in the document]
adisos: ok, I think that the dict key should be consistent and not depend on the config name, so maybe change to `conns_config1`
and `conns_config2` keys instead.

nca/Utils/QueryOutputHandler.py
```diff
@@ -0,0 +1,395 @@
...
+        :param explanation_description: the relevant description of this output explanation
+        :rtype: str
+        """
+        result = []
+        delimiter = ' '
+        for record in self.policies_pods:
+            result.append(f'{record.first_policy.split(delimiter)[0]}_1: {record.first_policy.split(delimiter)[1]}, '
+                          f'{record.second_policy.split(delimiter)[0]}_2: {record.second_policy.split(delimiter)[1]},'
+                          f' pods: {record.common_pods}')
+        return explanation_description + ':\n' + '\n'.join(result)
```
adisos, line 114: can we simplify this code? 
a few things to consider:
- a function that gets an input policy title and returns its components (policy name, policy kind,..)
- hold common pods as PeerSet or list of peer names, to avoid splitting by some character that could be changed in the future.
- missing documentation of the expected format for policy_str (if you are relying on this in the code it should be documented).

adisos (changes requested):
Hi @shireenf-ibm , added another comment.
Also, consider if `QueryOutputHandler.py` should be in NetworkConfig rather than in Utils. It is very specifically related to generated outputs of network config query.

adisos (changes requested):
Added a few more comments.

nca/NetworkConfig/QueryOutputHandler.py
```diff
@@ -0,0 +1,397 @@
+#
+# Copyright 2020- IBM Inc. All rights reserved
+# SPDX-License-Identifier: Apache2.0
+#
+from __future__ import annotations
```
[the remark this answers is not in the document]
shireenf-ibm: removing this import (after reordring the file classes and removing `CombinedExplanation`) , worked successfully on my PyCharm but failed on github (still trying to understand why), so I had to return it.

nca/NetworkConfig/QueryOutputHandler.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        returns dict object of policies to ingress/egress rules and its description
+        :param str xgress_description: the relevant description of this oof the given rules' dict
+        :param str prefix: indicates if rules types are ingress or egress
+        :return dict of description and pairs of policy and rules indexes
+        :rtype dict
+        """
+        rules = []
+        for key, value in xgress_rules_dict.items():
+            rules.append({'policy': key.split()[1], prefix + '_rules_indexes': [str(idx) for idx in value]})
```
adisos, line 93: I would add a method to extract policy name from title instead of writing directly `key.split()[1]` , as it is less clear.

nca/NetworkConfig/QueryOutputHandler.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        returns self type of OutputExplanation arranged in dict/s objects
+        a dict for each item in self with its relevant description
+        :param explanation_description: the relevant description of this output explanation
+        :rtype: list[dict]
+        """
+        result = []
+        if self.policies_list:
+            result.append({'description': self._get_policies_description(explanation_description),
+                           'policies': [policy.split()[1] for policy in self.policies_list]})
```
adisos, line 106: same with this `split`.

adisos (changes requested):

nca/NetworkConfig/QueryOutputHandler.py
```diff
@@ -0,0 +1,397 @@
...
+        computes and returns the output explanation is str, so it may be used to write the query answer in txt format.
+        An OutputExplanation object may have only one field representing the explanation data besides to
+         explanation_description, so the output is computed by calling the get_explanation_in_str method of this field
+         :return the explanation written with its description in txt format
+         :rtype: str
+        """
+        for val in self.__dict__.values():
+            if not isinstance(val, str) and val is not None:
+                return val.get_explanation_in_str(self.explanation_description)
+
```
adisos, line 317: I think the function `get_explanation_in_str` is very dependent on the current fields of `OutputExplanation`, and this could be error-prone if someone in the future adds other fields to this class.
One option I could think of is having a base class `OutputExplanation` only with `explanation_description`, and letting the other classes (such as `IntersectPodsExplanation`) inherit from this class. You could also add a wrapper class for the `str_explanation`.
What do you think?

shireenf-ibm: yes, I agree , I'll try to re-organize the code with inheritance

Issue #351, opened 2022-11-07
support scheme file with missing global section


adisos:
When global section is missing, can try to access live cluster if exists, and if not - use empty global config (and then avoid error of access to live cluster).

PR #354, opened 2022-11-10
ignore errors to communicate with live cluster if it is not an explicit requirement

shireenf-ibm:
#351
any implicit attempt to load from k8s live cluster will occur silently
explicit attempts (e.g. k8s is an input resource) - runs as usual (communication errors are not ignored)

### DIFF
--- a/nca/Utils/CmdlineRunner.py
+++ b/nca/Utils/CmdlineRunner.py
@@ -21,18 +21,19 @@ class CmdlineRunner:
     ignore_live_cluster_err = False
 
     @staticmethod
-    def run_and_get_output(cmdline_list):
+    def run_and_get_output(cmdline_list, helm_flag=False):
         """
         Run an executable with specific arguments and return its output to stdout
         if a communicate error occurs, it will be ignored in case this is a silent try to communicate with live cluster,
         otherwise, will be printed to stderr
         :param list[str] cmdline_list: A list of arguments, the first of which is the executable path
+        :param helm_flag: indicates if the executable is helm - communicate errors are always considered for helm
         :return: The executable's output to stdout ( a list-resources on success, otherwise empty value)
         :rtype: str
         """
         cmdline_process = subprocess.Popen(cmdline_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         out, err = cmdline_process.communicate()
-        if err and not CmdlineRunner.ignore_live_cluster_err:
+        if err and (not CmdlineRunner.ignore_live_cluster_err or helm_flag):
             print(err.decode().strip('\n'), file=sys.stderr)
         return out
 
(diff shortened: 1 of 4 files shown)


nca/Utils/CmdlineRunner.py
```diff
@@ -9,23 +9,32 @@
...
         :param list[str] cmdline_list: A list of arguments, the first of which is the executable path
-        :return: The executable's output to stdout
+        :return: The executable's output to stdout ( a list-resources on success, otherwise empty value)
         :rtype: str
         """
-        cmdline_process = subprocess.Popen(cmdline_list, stdout=subprocess.PIPE)
-        return cmdline_process.communicate()[0]
+        cmdline_process = subprocess.Popen(cmdline_list, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, err = cmdline_process.communicate()
+        if err and not CmdlineRunner.ignore_live_cluster_err:
```
adisos, line 35: This function (`run_and_get_output`) gets called also by `resolve_helm_chart`, and on this case the flag `ignore_live_cluster_err` is not relevant.

nca/Utils/CmdlineRunner.py
```diff
@@ -9,23 +9,33 @@
...
 class CmdlineRunner:
     """
     A stateless class with only static functions to easily get k8s and calico resources using kubectl and calicoctl
     """
+    # a static variable to indicate if we want to ignore errors from running executable command
+    ignore_live_cluster_err = False
+
     @staticmethod
-    def run_and_get_output(cmdline_list):
+    def run_and_get_output(cmdline_list, helm_flag=False):
```
adisos, line 24: I think a better approach would be to set a k8s live cluster flag (default as False, to be set true when called from the relevant functions). Since `run_and_get_output` might be extended for other use cases in the future.

nca/Utils/CmdlineRunner.py
```diff
@@ -95,7 +98,7 @@ def get_k8s_resources(resource):
         cmdline_list = ['kubectl', 'get', resource, '-o=json']
         if resource in ['networkPolicy', 'authorizationPolicy', 'pod', 'ingress', 'Gateway', 'VirtualService', 'sidecar']:
             cmdline_list.append('--all-namespaces')
-        return CmdlineRunner.run_and_get_output(cmdline_list)
+        return CmdlineRunner.run_and_get_output(cmdline_list, check_for_silent_exec=True)
```
adisos, line 101: Shouldn't also be set True from `get_calico_resources`, not only from `get_k8s_resources `?

shireenf-ibm: currently if the topology section is missing / or  user runs cmdline without resources or without policies - our default is to load missing resources from k8s live cluster - so it is the only relevant case for a silent attempt

we load from calico only if the input is calico - a provided "live cluster" input does not ignore errors for any of the layers

adisos (approved):

PR #363, opened 2022-11-22
No longer using requirements txt

### LINKED ISSUE #362: Avoid duplicating dependencies in requirements.txt and setup.cfg
It seems as if we can live without `requirements.txt`.
The procedure for creating a development environment would be on Windows:
```shell
python -m venv venv
source venv/Scripts/activate
pip install .
```
On Linux:
```shell
virtualenv venv
source venv/bin/activate
pip install .
```

zivnevo:
Fixes #362

### DIFF
--- a/.github/actions/setup-nca-env/action.yaml
+++ b/.github/actions/setup-nca-env/action.yaml
@@ -8,5 +8,5 @@ runs:
       python-version: 3.8
       architecture: x64
       cache: pip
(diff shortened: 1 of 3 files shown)


zivnevo:
For now, closing this PR, as this seems to have more disadvantages than advantages.
Will reconsider later, possibly using tools like Poetry.

PR #364, opened 2022-11-22
Small fixes for master

tanyaveksler:


adisos (commented):
@tanyaveksler I think we should split this to separate branches/PRs and open relevant issues, explaining each issue what should be fixed.

tanyaveksler:
Instead create three more specific PRs.

PR #370, opened 2022-11-27
Fixed peer dimension in hyper cube set

tanyaveksler:
Better handling sorted_peer_list;
Fixing calculation of PeerSet from interval set;
More accurate assertions.

Signed-off-by: Tanya <tatyana@il.ibm.com>


adisos (changes requested):
I think we should add some documentation: why peer_set is a special dimension, where in the flow it is expected to change, why it is sometimes set or not, how it is handled for comparison, etc.

tanyaveksler:
This pull request is out of date. Its changes are covered by Optimized_HC_set branch.

PR #378, opened 2022-12-04
Issue 336 output query - json support

shireenf-ibm:
Committed:
1. json support for queries that are handled by `QueryOutputHandler `
2. json support for `ConnectivityMapQuery` and `SemanticDiffQuery `
3. changes in `NetworkConfigQueryRunner.py` :
	1. to support writing a json "file".
	single json file should always have one top level item, 
	since we run queries on `NetworkConfigList` and hence may run number of iterations on same query, I had to store json results (of each iteration) in a list
	and after all iterations are done to dump it back into string - so we get a fixed json object that includes all results under one top level 
	(`example with multiple results: equiv_all_range1_json, tests/calico_testcases/example_policies/testcase16-nets/testcase16-scheme.yaml`)
		
	2. made minor changes to avoid code repetition in methods that run the query iterations

4. json examples (for all supported queries and expected outputs)


nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -117,59 +119,115 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
+        :param Union[list, str] output: the output of running all query iterations
+        :param int queries_not_executed: number of times query was not executed
+        :return the results: numerical result, output - str , num of not executed
+        :rtype: int, str, int
+        """
+        if self.output_configuration.outputFormat == 'json':
+            if len(output) == 1:
+                output = self._dump_json_output(output[0])
+            else:
+                output = self._dump_json_output(output)
```
adisos, line 224: I think the output should be consistent as a list of objects in both cases.

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -117,59 +119,115 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
         return res, output, queries_not_executed
 
+    @staticmethod
+    def _dump_json_output(output):
+        return json.dumps(output, sort_keys=False, indent=2)
+
```
adisos, line 230: This method is redundant. You can use `JsonOutputHandler.dump_content` instead.

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -117,59 +119,115 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
+        initializes the query variable results as following
+        res : 0 (the numerical result)
+        output : will be initialized to an empty list if the output format is json, otherwise, empty str
+        (the formatted explanation result of the query)
+        queries not executed : 0 (the number of not executed queries)
+        rtype: int , Union[list, str], int
+        """
+        # json results will be appended in a list, and finally be dumped into string to
+        # ensure all results are written under one top level object to get a fixed json output format
+        output = [] if self.output_configuration.outputFormat == 'json' else ''
```
adisos, line 180: I think we should prefer holding the same types of objects.
We can hold a list of outputs and eventually use strings join by newline or json dump , depending on the output format.

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -117,59 +119,115 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
+        :param int iter_res: the numerical result from running this iteration
+        :param Union[list, str] output: the output from previous iterations
+        :param str iter_output: the query output from last iteration
+        :param int num_not_exec: the number of not executed queries from previous iterations
+        :param int iter_not_exec: the query_not_executed result from last iteration (0/1)
+        rtype: int , Union[list, str], int
+        """
+        result += iter_res
+        if self.output_configuration.outputFormat == 'json':
+            output.append(json.loads(iter_output))
```
adisos, line 202: so for the json case , we get the output in string(from json) format, then convert it back into a dict to be appended to to the list, and eventually use conversion back to string of the entire list of dict objects? 
Can we instead return the output in the dict object originally ? (could also be relevant for yaml) so we don't have to use` json.loads` here?

adisos (changes requested):
Added a few comments.

nca/NetworkConfig/NetworkConfigQueryRunner.py
shireenf-ibm: Hi, I am making relevant fixes,
I want to change the format of yaml output to be  a "list" of query results , instead of multiple yaml objects separated with "---" 
e.g.:
instead of :
> 
```
query: equiv_all_range1_yaml
configs:
- equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
- equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-nets
numerical_result: 0
textual_result: equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
  and equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-nets
  are semantically equivalent.
---

query: equiv_all_range1_yaml
configs:
- equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
- equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-notNets
numerical_result: 0
textual_result: equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
  and equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-notNets
  are semantically equivalent.
---
```

the result will be:

```
- query: equiv_all_range1_yaml
  configs:
  - equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
  - equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-nets
  numerical_result: 0
  textual_result: equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
    and equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-nets
    are semantically equivalent.

- query: equiv_all_range1_yaml
  configs:
  - equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
  - equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-notNets
  numerical_result: 0
  textual_result: equiv-ranges-writing-games/kube-system/testcase16-simple-all-range
    and equiv-ranges-writing-games/kube-system/testcase16-nets-all-range-partition2-notNets
    are semantically equivalent.
```

I want to know if it is okay for you too?

this will also be consistent with ConnectivityMap and SemanticDiff yaml results - 
and then I can dump results of all queries only once at the end in the NetworkConfigQueryRunner.py

adisos: yes, looks good to me.

adisos: @zivnevo can you also confirm this output change is ok?

shireenf-ibm:
Hi @adisos , @zivnevo ,
I have a suggestion, 
if it is possible and doesn't affect other apps (if any use it), I want to make minor changes to the yaml output of `ConnectivityMapQuery `and `SemanticDiffQuery` to be consistent with the yaml format of other queries.
This way, yaml and txt output of all queries will be handled in the `QueryOutputHandler.py` file 
This will not affect the other output formats that these two queries support.

example of the changes:
image
(we may omit description here)

image

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -117,57 +120,105 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
+        rtype: int , list, int
+        """
+        result += iter_res
+        output.append(iter_output)
+        num_not_exec += iter_not_exec
+        return result, output, num_not_exec
+
+    def _return_final_query_results(self, res, output, queries_not_executed):
+        """
+        gets the final query results after running it on all iterations of configs_array
```
adisos, line 205: These methods are used for a repeated set of objects, both as input and output (`res, output, queries_not_executed)`.
I think it might be easier to encapsulate this set of objects in another class, that can support the required operations (e.g. method such as 'add'.)

nca/FWRules/MinimizeFWRules.py
```diff
@@ -676,7 +675,9 @@ def _get_fw_rules_content_str(self, query_name, req_format, add_txt_header, add_
         :param req_format: a string of the required format, should be in FWRule.supported_formats
         :param add_txt_header:  bool flag to indicate if header of fw-rules query should be added in txt format
         :param add_csv_header: bool flag to indicate if header csv should be added in csv format
-        :return: a string of the query name + fw-rules in the required format
+        :return: a dict of the fw-rules if the required format is json or yaml, else
+        a string of the query name + fw-rules in the required format
+        :rtype: Union[str, dict]
```
adisos, line 680: change the function name, since it does not always return `str` anymore.

adisos (changes requested):
Added a few more comments.

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -120,106 +166,38 @@ def _execute_pair_configs_query(self, query_type, config1, config2, cmd_line_fla
...
-        res, output, queries_not_executed = self._init_query_results_values()
+        query_result = QueryResult()
         for config in self.configs_array:
-            query_res, query_output, query_not_executed =\
-                self._execute_one_config_query(self.query_name, self._get_config(config))
-            res, output, queries_not_executed = \
-                self._update_query_results_after_one_iteration(res, query_res, output, query_output,
-                                                               queries_not_executed, query_not_executed)
-        return self._return_final_query_results(res, output, queries_not_executed)
+            query_result.update(self._execute_one_config_query(self.query_name, self._get_config(config)))
```
adisos, line 171: Maybe the function `_execute_one_config_query` can also return a `QueryResult` object?
The input arg to `QueryResult.update()` can be a `QueryResult` object rather than a plain tuple.

shireenf-ibm: the tuple in the `QueryResult.update()` is actually returned from `NetworkConfigQuery.execute_and_compute_output_in_required_format`
(which is called and returned by `_execute_one_config_query `and _`execute_pair_configs_query `from this file).

it is not exactly the same as the `QueryResult` since:
1- the tuple includes a query result from one iteration while `QueryResult'`s goal is to include the "sum" of all iterations till now
2 - the tuple's "query_output" variable is `Union[str, dict]`, while the query_output of the `QueryResult` is  `list `that stores results from all different "iterations" 
3- while it is possible to add a new member to the `QueryResult `(output_from_one_iteration), it will be problematic to import it in the `NetworkConfigQuery ` to be returned (this will cause a circular import , since this file imports all from  `NetworkConfigQuery` )
and I think that it is not right to move the `QueryResult` class to the `NetworkConfigQuery.py `since it is used for calculations that are relevant only in this file (i.e. `NetworkConfigQueryRunner.py`))| (do you agree?)

So another approach to avoid using a tuple here, is to add another `dataClass ` in `NetworkConfigQuery.py` which will consist of the  returned values in `NetworkConfigQuery.execute_and_compute_output_in_required_format`
does this sounds good?

adisos: ok, so let's just keep it as it is now.

adisos (approved):
looks good, ok to merge.

PR #390, opened 2023-01-03
1. add --debug flag

haim-kermany:
2. catch and pring exceptions + their traces

Signed-off-by: haim-kermany <haim.kermany@gmail.com>

### DIFF
--- a/nca/nca_cli.py
+++ b/nca/nca_cli.py
@@ -7,6 +7,7 @@ import argparse
 import time
 import os
 import sys
+import traceback
 from pathlib import Path
 from nca.Utils.OutputConfiguration import OutputConfiguration
 from nca.NetworkConfig.NetworkConfigQueryRunner import NetworkConfigQueryRunner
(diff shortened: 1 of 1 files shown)


nca/nca_cli.py
```diff
@@ -324,11 +326,17 @@ def nca_main(argv=None):
...
-
-    _do_every(args.period * 60, run_args, args)
+    try:
+        if args.period <= 0:
+            ret_val = run_args(args)
+            return 0 if args.return_0 else ret_val
+
+        _do_every(args.period * 60, run_args, args)
+    except Exception as e:
+        print('Error: {}'.format(e))
```
zivnevo, line 336:
```suggestion
        print(f'Error: {e}')
```
We always use f-strings in this project

zivnevo (changes requested):
Just this one tiny change

nca/nca_cli.py
haim-kermany: done

PR #397, opened 2023-01-08
issue 384 - error on empty peer container 

haim-kermany:
Signed-off-by: haim-kermany <haim.kermany@gmail.com>

### DIFF
--- a/nca/SchemeRunner.py
+++ b/nca/SchemeRunner.py
@@ -198,12 +198,11 @@ class SchemeRunner(GenericYamlParser):
                 if res != expected:
                     self.warning(f'Unexpected result for query {query_name}: Expected {expected}, got {res}\n', query)
                     self.global_res += 1
-            if 'expectedNotExecuted' in query:
-                expected_not_executed = query['expectedNotExecuted']
-                if not_executed != expected_not_executed:
-                    self.warning(f'{query_name} was not executed {not_executed} times. '
-                                 f'Although, expected to not be executed {expected_not_executed} times')
-                    self.global_res += 1
+            expected_not_executed = query.get('expectedNotExecuted', 0)
+            if not_executed != expected_not_executed:
+                self.warning(f'{query_name} was not executed {not_executed} times. '
+                             f'Although, expected to not be executed {expected_not_executed} times')
+                self.global_res += 1
             if comparing_err != 0:
                 self.warning(f'Unexpected output comparing result for query {query_name} ')
                 self.global_res += 1
--- a/nca/SchemeRunner.py
+++ b/nca/SchemeRunner.py
@@ -200,8 +200,10 @@ class SchemeRunner(GenericYamlParser):
                     self.global_res += 1
             expected_not_executed = query.get('expectedNotExecuted', 0)
             if not_executed != expected_not_executed:
-                self.warning(f'{query_name} was not executed {not_executed} times. '
-                             f'Although, expected to not be executed {expected_not_executed} times')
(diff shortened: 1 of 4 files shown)


adisos (commented):
Maybe we should consider using the mechanism of `query_not_executed` instead of raising an exception? 
@zivnevo @shireenf-ibm  what do you think?

shireenf-ibm:
> Maybe we should consider using the mechanism of `query_not_executed` instead of raising an exception? @zivnevo @shireenf-ibm what do you think?

I agree, empty container means that the query is not executed. (this will also cause the scheme/cmdline to fail)
If you want to "attach" the "explaining message" to the result, then in my opinion the check should be done in the `execute`, so the message may be written in the query_answer.output_result

shireenf-ibm:
suggestions TBD:
1. the error message should be written to the query_answer.output_result, so the result would be printed correctly in any required format, my suggestion is to move the "empty container check" to the "execute" method so the values are returned within QueryAnswer object
2. currently, for scheme files, when a query is not executed, tests will fail (return with exit code > 0 ) only if the user states that "    expectedNotExecuted: 1"  in the scheme file , otherwise the not_executed value is ignored, this should be changed in the SchemeRunner.py ( not executed value should be checked anyway and if > 0 to be added to the global res)

haim-kermany:
> suggestions TBD:
> 
>     1. the error message should be written to the query_answer.output_result, so the result would be printed correctly in any required format, my suggestion is to move the "empty container check" to the "execute" method so the values are returned within QueryAnswer object
> 
>     2. currently, for scheme files, tests will fail (return with exit code > 0 ) only if the user states that "expectedNotExecuted: 1" and it was actually executed (or opposite case)- i.e only if `expectedNotExecuted` is provided, otherwise the not_executed value is ignored,
>        this should be changed in the SchemeRunner.py ( not executed value should be checked anyway and if > 0 to be added to the global res, unless this is declared by the user)

implemented

nca/SchemeRunner.py
```diff
@@ -198,12 +198,11 @@ def run_queries(self, query_array):
...
-            if 'expectedNotExecuted' in query:
-                expected_not_executed = query['expectedNotExecuted']
-                if not_executed != expected_not_executed:
-                    self.warning(f'{query_name} was not executed {not_executed} times. '
-                                 f'Although, expected to not be executed {expected_not_executed} times')
-                    self.global_res += 1
+            expected_not_executed = query.get('expectedNotExecuted', 0)
+            if not_executed != expected_not_executed:
+                self.warning(f'{query_name} was not executed {not_executed} times. '
+                             f'Although, expected to not be executed {expected_not_executed} times')
```
adisos, line 204: I would consider changing the warning if `expectedNotExecuted` was not specified in the scheme file, to just include the text that describes the query was not executed.

adisos (changes requested):

tests/run_all_tests.py
```diff
@@ -323,6 +323,9 @@ def run_tests_spec(self, tests_spec):
                     continue
                 if self.test_files_spec.type == 'scheme' and not fnmatch(file, '*-scheme.yaml'):
                     continue
+                # todo: remove the followin if after fixing issue 396
+                if file == 'scan-dir-test-scheme.yaml':
+                    continue
```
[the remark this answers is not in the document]
haim-kermany: done

nca/SchemeRunner.py
haim-kermany: done

adisos (approved):
looks good

PR #412, opened 2023-01-23
cases to exclude ipv6 addresses from connectivity reports

shireenf-ibm:
Committed:
exclude ipv6 addresses from connectivity reports **unless** :
1. ipv6 addresses were referenced by config's policies
2. the user requested printing it using OutputConfiguration.connectivityExcludeIpv6Edges = false

**A question:**
- [x] Do we need to add a similar flag for cmdline too?

**Note:** 
If the resource should reference IP addresses and the parser fill them automatically if the relevant field is missing, still the parser will fill with ipv4 + ipv6 (and then ipv6 addresses will not be excluded because they are referenced by the policies), because:
1.we don't know if the user intended to have both 
2. this PR affects only ConnectivityMap Query reports, changing the parsers' default will affect other queries too


adisos:
> Committed: exclude ipv6 addresses from connectivity reports **unless** :
> 
>     1. ipv6 addresses were referenced by config's policies
> 
>     2. the user requested printing it using OutputConfiguration.connectivityExcludeIpv6Edges = false
> 
> 
> **A question:**
> 
>     * [ ]  Do we need to add a similar flag for cmdline too?
> 
> 
> **Note:** If the resource should reference IP addresses and the parser fill them automatically if the relevant field is missing, still the parser will fill with ipv4 + ipv6 (and then ipv6 addresses will not be excluded because they are referenced by the policies), because: 1.we don't know if the user intended to have both 2. this PR affects only ConnectivityMap Query reports, changing the parsers' default will affect other queries too
> 
> Signed-off-by: shireenf-ibm [shireenf@il.ibm.com](mailto:shireenf@il.ibm.com)

Isn't this relevant to other queries as well?

shireenf-ibm:
> > Committed: exclude ipv6 addresses from connectivity reports **unless** :
> > ```
> > 1. ipv6 addresses were referenced by config's policies
> > 
> > 2. the user requested printing it using OutputConfiguration.connectivityExcludeIpv6Edges = false
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **A question:**
> > ```
> > * [ ]  Do we need to add a similar flag for cmdline too?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **Note:** If the resource should reference IP addresses and the parser fill them automatically if the relevant field is missing, still the parser will fill with ipv4 + ipv6 (and then ipv6 addresses will not be excluded because they are referenced by the policies), because: 1.we don't know if the user intended to have both 2. this PR affects only ConnectivityMap Query reports, changing the parsers' default will affect other queries too
> > Signed-off-by: shireenf-ibm [shireenf@il.ibm.com](mailto:shireenf@il.ibm.com)
> 
> Isn't this relevant to other queries as well?

In this PR:
- I use the new flag only in the ConnectivityMap query since the issue asked to remove them from connectivity reports
- In Peer.py : the default for exclude_ipv6 is false , so it will exclude the ipv6 addresses only for connectivityMap

adisos (commented):
looks good, added a few comments.

- Can you add some documentation about the new flag? 

- Why are the following expected output files still include ipv6 addresses:

```
tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_dot.dot
tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_endpoints_deployments_dot.dot
tests/k8s_testcases/example_policies/expected_output/services1_connectivity_map.txt
```

shireenf-ibm:
> * Why are the following expected output files still include ipv6 addresses:
> 
> 
> ```
> tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_dot.dot
> tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_endpoints_deployments_dot.dot
> tests/k8s_testcases/example_policies/expected_output/services1_connectivity_map.txt
> ```

-  `tests/k8s_testcases/example_policies/expected_output/services1_connectivity_map.txt:`
this file is not used, the "real" expected output file is `tests/k8s_testcases/expected_output/services1_connectivity_map.txt`
and it does not include ipv6 results 
deleting the unused file 

- ```
tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_dot.dot
tests/fw_rules_tests/policies/expected_output/subset_deployment_no_subset_endpoints_deployments_dot.dot
```
I didn't find any references for both files in the *-scheme.yaml files -> so these files are unused either , removing them

File: nca/NetworkConfig/NetworkConfig.py [hunk 1]
```diff
@@ -172,17 +172,33 @@ def get_affected_pods(self, is_ingress, layer_name):
...
+        Excluding the IPv6 addresses will be enabled if the exclude_ipv6 param is True and
+        IPv6 addresses in all the policies of the config (if existed) were added automatically by the parser
+        and not referenced by user
+        :param bool exclude_ipv6: indicates if to exclude ipv_6 non-referenced addresses
+        :rtype bool
+        """
+        for policy in self.policies_container.policies.values():
+            if policy.has_ipv6_addresses:  # if at least one policy has referenced ipv6 addresses, ipv6 will be included
+                return False
+        return exclude_ipv6  # getting here means all policies didn't reference ipv6, excluding depends on the given flag
```
adisos, line 187: if the given flag `exclude_ipv6` is `false`, then no need to iterate the policies

adisos (approved):

Issue #410, opened 2023-01-23
Code for REST-API server should move to another repo


zivnevo:
The new repo will consume NCA as a library.
This will allow NCA to get rid of the dependency in Flask.

PR #413, opened 2023-01-25
Issue #410: Remove support for daemon mode

zivnevo:
Running NCA as a REST server now lives in its own repo: https://github.com/np-guard/nca-rest-server


shireenf-ibm (approved):
looks good.

PR #426, opened 2023-02-02
Issue 422 support jpg format

haim-kermany:
user can choose jpg as an output of the connectivity graph
a dot file is created, and converted to dot file


zivnevo (changes requested):
Also need to update `README.md`.
Maybe mention there that for a jpeg format, Graphviz must be installed and that the directory of its `dot` executable must be in PATH.

haim-kermany:
> Also need to update `README.md`. Maybe mention there that for a jpeg format, Graphviz must be installed and that the directory of its `dot` executable must be in PATH.

done

zivnevo (approved):

PR #447, opened 2023-02-26
Optimized hyper cube set

tanyaveksler:


nca/CoreDS/TcpLikeProperties.py
```diff
@@ -36,42 +38,49 @@ class TcpLikeProperties(CanonicalHyperCubeSet):
...
-                          "paths": {"value": paths,
-                                    "is_all": paths is None},
-                          "peers": {"value": peers,
-                                    "is_all": peers is None}}
+        dims_to_values = {"src_peers": {"value": src_peers, "is_all": src_peers is None},
+                          "dst_peers": {"value": dst_peers, "is_all": dst_peers is None},
+                          "protocols": {"value": protocols, "is_all": protocols is None or protocols.is_whole_range()},
+                          "src_ports": {"value": source_ports.port_set, "is_all": source_ports.is_all()},
+                          "dst_ports": {"value": dest_ports.port_set, "is_all": dest_ports.is_all()},
+                          "methods": {"value": methods, "is_all": methods is None or methods.is_whole_range()},
```
adisos, line 79: is it possible that `methods` is `None`? the default is `MethodSet(True)` and not `None` in the initialization

nca/CoreDS/TcpLikeProperties.py
```diff
@@ -36,42 +38,49 @@ class TcpLikeProperties(CanonicalHyperCubeSet):
...
+                          "hosts": {"value": hosts, "is_all": hosts is None},
+                          "paths": {"value": paths, "is_all": paths is None},
+                          "icmp_type": {"value": icmp_type, "is_all": icmp_type is None},
```
adisos, line 82: is it possible that `icmp_type` is an object with the range of the entire domain rather than `None`?

nca/CoreDS/TcpLikeProperties.py
```diff
@@ -145,10 +154,10 @@ def get_cube_dict(self, cube, is_txt=False):
             dim_domain = DimensionsManager().get_dimension_domain_by_name(dim)
             if dim_domain == dim_values:
                 continue  # skip dimensions with all values allowed in a cube
-            if dim == 'methods':
+            if dim == 'protocols' or dim == 'methods':
                 values_list = str(dim_values)
-            elif dim == "peers":
-                values_list = self.base_peer_set.get_peer_list_by_indices(dim_values)
+            elif dim == "src_peers" or dim == "dst_peers":
+                values_list = self.base_peer_set.get_peer_set_by_indices(dim_values)
```
adisos, line 160: shouldn't `values_list` be a list of `str` values?

adisos (commented):
hi @tanyaveksler , added a few initial comments to start with.

Currently `TcpLikeProperties` is used for both the original version and the optimized version.
Maybe it might make sense to separate it into another type, such as `ConnectivityCubes` for the optimized version ? and keep the original  `TcpLikeProperties` as it was?

adisos (commented):
additional comments

nca/FWRules/ConnectivityGraph.py
```diff
@@ -355,3 +460,99 @@ def conn_graph_has_fw_rules(self):
             if not conn:  # conn is "no connections":
                 return False
         return True
+
+
+class ConnectivityGraphOptimized(ConnectivityGraphPrototype):
```
adisos, line 465: remove `ConnectivityGraphOptimized`?

nca/FWRules/ConnectivityGraph.py
```diff
@@ -7,13 +7,40 @@
...
+from nca.CoreDS.ConnectionSet import ConnectionSet
+from nca.CoreDS.ProtocolSet import ProtocolSet
+from nca.CoreDS.TcpLikeProperties import TcpLikeProperties
 from .DotGraph import DotGraph
 from .MinimizeFWRules import MinimizeCsFwRules, MinimizeFWRules
 from .ClusterInfo import ClusterInfo
 
 
-class ConnectivityGraph:
+class ConnectivityGraphPrototype:
```
adisos, line 19: can remove `ConnectivityGraphPrototype` and use only `ConnectivityGraph`?

tanyaveksler: Done

nca/FWRules/ConnectivityGraph.py
tanyaveksler: Done

nca/CoreDS/TcpLikeProperties.py
tanyaveksler: Fixed. Also for protocols.

nca/CoreDS/TcpLikeProperties.py
tanyaveksler: Fixed, Also for icmp_code.

nca/CoreDS/TcpLikeProperties.py
tanyaveksler: It was actually a list of Peers, and I changed it to PeerSet. It does not matter, it is further converted to str.

nca/CoreDS/TcpLikeProperties.py
```diff
@@ -317,6 +366,271 @@ def _get_first_item_str(self):
...
+            if dim_name == "src_peers" or dim_name == "dst_peers":
+                return PeerSet()
+            elif dim_name == "src_ports" or dim_name == "dst_ports":
+                return PortSet()
+            elif dim_name == "protocols":
+                return ProtocolSet()
+            elif dim_name == "methods":
+                return MethodSet()
+            else:
+                return None
```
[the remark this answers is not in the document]
tanyaveksler: Actually, it should be full rather than empty. Since different dimensions have different full domain values (and we probably do not want to return here all these values), I changed it to None for all non active dimensions.
Currently, this function is used only for "src_peers" and "dst_peers" dimensions.

nca/CoreDS/TcpLikeProperties.py
```diff
@@ -317,6 +366,271 @@ def _get_first_item_str(self):
...
+        icmp_code = cube_dict_copy.pop("icmp_code", None)
+        assert not cube_dict_copy
+        return TcpLikeProperties.make_tcp_like_properties(peer_container, src_ports=src_ports, dst_ports=dst_ports,
+                                                          protocols=protocols, src_peers=src_peers, dst_peers=dst_peers,
+                                                          paths_dfa=paths_dfa, hosts_dfa=hosts_dfa, methods=methods,
+                                                          icmp_type=icmp_type, icmp_code=icmp_code)
+
+    @staticmethod
+    def make_empty_properties(peer_container=None):
+        return TcpLikeProperties(base_peer_set=peer_container.peer_set.copy() if peer_container else None, create_empty=True)
```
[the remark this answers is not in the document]
tanyaveksler: It's a copy of PeerSet, not PeerContainer. It's not efficient to copy it inside the init, because make_tcp_like_properties() calls to init multiple times.

tanyaveksler:
> 
I think that maintaining the original TcpLikeProperties while developing in parallel a new extended ConnectivityCubes data structure would be very complicated and unreadable for the code review, it would involve copy-past of many many fragments of code.
Also, if we are planning to eventually move to the optimized solution, this effort will not be worth timewise.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+        if (not src_ports.named_ports or not src_peers) and (not dst_ports.named_ports or not dst_peers):
+            # Should not resolve named ports
+            return ConnectivityProperties(source_ports=src_ports, dest_ports=dst_ports,
+                                          protocols=protocols, methods=methods, paths=paths_dfa, hosts=hosts_dfa,
+                                          icmp_type=icmp_type, icmp_code=icmp_code,
+                                          src_peers=src_peers_interval, dst_peers=dst_peers_interval,
+                                          base_peer_set=base_peer_set)
+        # Resolving named ports
+        conn_properties = ConnectivityProperties.make_empty_props(peer_container)
+        if src_ports.named_ports and dst_ports.named_ports:
```
adisos, line 447: what about `excluded_named_ports` of `PortSet` objects? they should also be handled and resolved.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+        src_ports = cube_dict_copy.pop("src_ports", PortSet(True))
+        dst_ports = cube_dict_copy.pop("dst_ports", PortSet(True))
+        protocols = cube_dict_copy.pop("protocols", ProtocolSet(True))
+        src_peers = cube_dict_copy.pop("src_peers", None)
+        dst_peers = cube_dict_copy.pop("dst_peers", None)
+        paths_dfa = cube_dict_copy.pop("paths", None)
+        hosts_dfa = cube_dict_copy.pop("hosts", None)
+        methods = cube_dict_copy.pop("methods", MethodSet(True))
+        icmp_type = cube_dict_copy.pop("icmp_type", DimensionsManager().get_dimension_domain_by_name('icmp_type'))
+        icmp_code = cube_dict_copy.pop("icmp_code", DimensionsManager().get_dimension_domain_by_name('icmp_code'))
```
adisos, line 522: we have many places (at least 3) where we duplicate the hard-coded semantics of defaults for a missing dimension in a cube input. can we avoid this duplication?

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +570,114 @@ def get_non_tcp_connections():
         res.add_all_connections([ProtocolNameResolver.get_protocol_number('TCP')])
         return res
         # return ConnectionSet(True) - ConnectionSet.get_all_TCP_connections()
+
+    # TODO - after moving to the optimized HC set implementation,
+    #  get rid of ConnectionSet and move the code below to ConnectivityProperties.py
+    @staticmethod
+    def tcp_properties_to_fw_rules(tcp_props, cluster_info, peer_container, ip_blocks_mask,
+                                   connectivity_restriction):
+        """
```
adisos, line 579: update the renaming of `tcp_props` to `conn_props`

nca/FWRules/ConnectivityGraph.py
```diff
@@ -384,6 +448,39 @@ def get_connectivity_dot_format_str(self, connectivity_restriction=None):
                 dot_graph.add_edge(src_name=edge[0][0], dst_name=edge[1][0], label=conn_str, is_dir=False)
         return dot_graph.to_str()
 
+    def convert_to_tcp_like_properties(self, peer_container):
```
adisos, line 451: update renaming of `tcp_props` to `conn_props`

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+        """
+        if dim_name not in self.active_dimensions:
+            return None
+        res = None
+        for cube in self:
+            cube_dict = self.get_cube_dict_with_orig_values(cube)
+            values = cube_dict.get(dim_name)
+            if values:
+                res = (res | values) if res else values
+        return res
```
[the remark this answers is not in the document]
tanyaveksler: Fixed the documentation

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+    def project_on_one_dimension(self, dim_name):
+        """
+        Build the projection of self to the given dimension
+        :param str dim_name: the given dimension
+        :return: the projection on the given dimension, having that dimension type (either IntervalSet or DFA)
+         or None if the given dimension is not active
+        """
+        if dim_name not in self.active_dimensions:
+            return None
+        res = None
```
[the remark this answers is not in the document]
tanyaveksler: Fixed.

nca/CoreDS/Peer.py
```diff
@@ -468,9 +478,21 @@ class PeerSet(set):
     Note #2: __contains__ is implemented under the assumption that item arg is from disjoint_ip_blocks()
     """
 
+    ipv4_highest_number = int(ip_network('0.0.0.0/0').broadcast_address)
+    ipv6_highest_number = int(ip_network('::/0').broadcast_address)
+    pod_highest_number = 9999  # assuming maximum 10,000 pods
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+        if src_peers:
+            base_peer_set |= src_peers
+            src_peers_interval = base_peer_set.get_peer_interval_of(src_peers)
+        else:
+            src_peers_interval = None
+        if dst_peers:
+            base_peer_set |= dst_peers
+            dst_peers_interval = base_peer_set.get_peer_interval_of(dst_peers)
+        else:
+            dst_peers_interval = None
```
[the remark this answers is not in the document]
tanyaveksler: You are right. Fixed.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+                peers_for_named_ports = dst_peers
+            assert peers_for_named_ports
+            assert not port_set_with_named_ports.port_set
+            assert len(port_set_with_named_ports.named_ports) == 1
+            port = list(port_set_with_named_ports.named_ports)[0]
+            for peer in peers_for_named_ports:
+                real_ports = ConnectivityProperties.resolve_named_port(port, peer, protocols)
+                if not real_ports:
+                    continue
+                if src_ports.named_ports:
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+            dst_peers_interval = base_peer_set.get_peer_interval_of(dst_peers)
+        else:
+            dst_peers_interval = None
+        if (not src_ports.named_ports or not src_peers) and (not dst_ports.named_ports or not dst_peers):
+            # Should not resolve named ports
+            return ConnectivityProperties(source_ports=src_ports, dest_ports=dst_ports,
+                                          protocols=protocols, methods=methods, paths=paths_dfa, hosts=hosts_dfa,
+                                          icmp_type=icmp_type, icmp_code=icmp_code,
+                                          src_peers=src_peers_interval, dst_peers=dst_peers_interval,
+                                          base_peer_set=base_peer_set)
```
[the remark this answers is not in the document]
tanyaveksler: Updated documentation

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+        # Resolving named ports
+        conn_properties = ConnectivityProperties.make_empty_props(peer_container)
```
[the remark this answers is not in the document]
tanyaveksler: Fixed

nca/CoreDS/ConnectivityProperties.py
tanyaveksler: excluded_named_ports should not appear in ConnectivityProperties together with dst_peers, because:
in the original implementation dst_peers are not a part of ConnectivityProperties and so cannot be resolved in ConnectivityProperties.
In the optimized implementation named ports are resolved as soon as ConnectivityProperties have been constructed, so excluded_named_ports are never created.
Added an assertion to ensure this.

nca/FWRules/ConnectivityGraph.py
tanyaveksler: Updated

nca/CoreDS/ConnectionSet.py
tanyaveksler: Updated

nca/CoreDS/ConnectivityProperties.py
tanyaveksler: Generalized.

nca/CoreDS/TcpLikeProperties.py
adisos: I think that for consistency considerations it may be better to change to list of `str` values as well.

nca/CoreDS/ConnectivityProperties.py
adisos: It is still duplicated in the `__init__` and `make_conn_props` default values for each argument, as well as in the `get_full_dim_having_orig_type` implementation.

nca/FWRules/ConnectivityGraph.py
adisos: this function name is still `convert_to_tcp_like_properties`

nca/CoreDS/ConnectivityProperties.py
zivnevo: How about storing a constant dict with all defaults and use dict.update() to override non-default values?

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,587 @@
...
+from .CanonicalHyperCubeSet import CanonicalHyperCubeSet
+from .DimensionsManager import DimensionsManager
+from .PortSet import PortSet
+from .MethodSet import MethodSet
+from .ProtocolSet import ProtocolSet
+from .Peer import PeerSet
+from .ProtocolNameResolver import ProtocolNameResolver
+
+
+class ConnectivityProperties(CanonicalHyperCubeSet):
```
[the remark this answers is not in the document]
adisos: still missing some documentation about how this class is used for the optimized use case vs the original one.

nca/CoreDS/ConnectivityProperties.py
adisos: I meant that it should be resolved while constructing the `ConnectivityProperties` object.
Adding another comment in `CalicoPolicyYamlParser.py` where this is relevant as an example.

nca/Parsers/CalicoPolicyYamlParser.py
```diff
@@ -453,33 +482,65 @@ def _parse_xgress_rule(self, rule, is_ingress, policy_selected_eps, is_profile):
...
+                        src_num_port_set = PortSet()
+                        src_num_port_set.port_set = src_res_ports.port_set.copy()
+                        dst_num_port_set = PortSet()
+                        dst_num_port_set.port_set = dst_res_ports.port_set.copy()
+                        conn_props = ConnectivityProperties.make_conn_props(self.peer_container,
+                                                                            src_ports=src_num_port_set,
+                                                                            dst_ports=dst_num_port_set,
+                                                                            protocols=protocols,
+                                                                            src_peers=src_res_pods,
+                                                                            dst_peers=dst_res_pods)
```
adisos, line 508: here `dst_res_port` can be a `PortSet` with `excluded_named_ports` not empty. (for instance if the input policy had 'notPorts: a'

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,634 @@
...
+        """
+        return ConnectivityProperties(create_empty=True)
+
+    @staticmethod
+    def make_all_props():
+        """
+        Returns all connectivity properties, representing logical True
+        :return: ConnectivityProperties
+        """
+        return ConnectivityProperties()
```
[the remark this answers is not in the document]
adisos: Is it valid to create `all_props` with an empty `base_peer_set`?

nca/FWRules/ConnectivityGraph.py
adisos: this function is not in use?

nca/CoreDS/ICMPDataSet.py
```diff
@@ -56,6 +56,23 @@ def get_properties_obj(self):
...
+        """
+        cube_dict = {}
+        for i, dim in enumerate(self.active_dimensions):
+            dim_values = cube[i]
+            dim_domain = DimensionsManager().get_dimension_domain_by_name(dim)
+            if dim_domain == dim_values:
+                continue  # skip dimensions with all values allowed in a cube
+            cube_dict[dim] = dim_values
+        return cube_dict
+
```
adisos, line 75: this function is not in use?

nca/Parsers/IstioPolicyYamlParser.py
```diff
@@ -485,30 +489,41 @@ def parse_ingress_rule(self, rule):
...
         # currently parsing only ports
         # TODO: extend operations parsing to include other attributes
+        conn_props = ConnectivityProperties.make_empty_props()
         if to_array is not None:
             connections = ConnectionSet()
             for operation_dict in to_array:
-                connections |= self.parse_operation(operation_dict)
+                conns = self.parse_operation(operation_dict)
+                connections |= conns
+                conn_props |= conns.convert_to_connectivity_properties(self.peer_container)
```
adisos, line 498: if the plan is to remove `ConnectionSet` after moving to the optimized reasoning, then this dependency in `conn_props |= conns.convert_to_connectivity_properties(self.peer_container)` should also be removed.

nca/Parsers/IstioPolicyYamlParser.py
```diff
@@ -485,30 +489,41 @@ def parse_ingress_rule(self, rule):
...
                     res_peers &= condition_res
                 elif isinstance(condition_res, ConnectionSet):
-                    connections &= condition_res
-
+                    condition_conns &= condition_res
         if not res_peers:
             self.warning('Rule selects no pods', rule)
-
-        return IstioPolicyRule(res_peers, connections)
+        condition_props = condition_conns.convert_to_connectivity_properties(self.peer_container)
```
adisos, line 518: same `ConnectionSet` dependency issue here

nca/CoreDS/ConnectivityProperties.py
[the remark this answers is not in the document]
tanyaveksler: Elaborated about it

nca/FWRules/ConnectivityGraph.py
tanyaveksler: Right, it's not in use. Removed

nca/CoreDS/TcpLikeProperties.py
tanyaveksler: Changed

nca/CoreDS/ConnectivityProperties.py
tanyaveksler: Right. Fixed dealing with `excluded_named_ports `- maintaining them in `ConnectivityCube ` and resolving when constructing `ConnectivityProperties `object.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,634 @@
...
+            return ProtocolSet(True)
+        if dim == "src_ports" or dim == "dst_ports":
+            return PortSet(True)
+        if dim == "methods":
+            return MethodSet(True)
+        if dim == "hosts" or dim == "paths":
+            return DimensionsManager().get_dimension_domain_by_name(dim)
+        if dim == "icmp_type" or dim == "icmp_code":
+            return DimensionsManager().get_dimension_domain_by_name(dim)
+        return None  # unknown dimension
```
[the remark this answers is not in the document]
tanyaveksler: Reorganized this using the new `ConnectivityCube `class.

nca/CoreDS/ConnectivityProperties.py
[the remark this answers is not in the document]
tanyaveksler: Added flag `create_all `to optimize `make_all_props`.
It is valid to create it with an empty `base_peer_set`, because the `src_peers `and `dst_peers `dimensions do not contain peers.

nca/Parsers/CalicoPolicyYamlParser.py
tanyaveksler: Fixed (as explained above).

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -663,59 +671,146 @@ def is_in_subset(self, peer):
...
+        print("Converting orig_conn_graph to tcp_like_properties...")
+        orig_tcp_props = orig_conn_graph.convert_to_connectivity_properties(self.config.peer_container)
+        assert orig_tcp_props.contained_in(opt_props) and opt_props.contained_in(orig_tcp_props)  # workaround for ==
+        # The following assert exposes the bug in HC set
+        assert not orig_tcp_props.contained_in(opt_props) or not opt_props.contained_in(orig_tcp_props) or \
+               orig_tcp_props == opt_props
+
+    def compare_fw_rules(self, fw_rules1, fw_rules2):
+        tcp_props1 = ConnectionSet.fw_rules_to_conn_props(fw_rules1, self.config.peer_container)
+        tcp_props2 = ConnectionSet.fw_rules_to_conn_props(fw_rules2, self.config.peer_container)
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ICMPDataSet.py
tanyaveksler: Actually, ICMPDataSet class is not longer used. Deleted it and its unit test.

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -663,59 +671,146 @@ def is_in_subset(self, peer):
...
-        res = QueryAnswer(True)
-        if self.output_config.outputFormat in ['json', 'yaml']:
-            res.output_explanation = [ComputedExplanation(dict_explanation=output_res)]
+    def compare_orig_to_opt_conn(self, orig_conn_graph, opt_props):
+        print("Converting orig_conn_graph to tcp_like_properties...")
+        orig_tcp_props = orig_conn_graph.convert_to_connectivity_properties(self.config.peer_container)
+        assert orig_tcp_props.contained_in(opt_props) and opt_props.contained_in(orig_tcp_props)  # workaround for ==
+        # The following assert exposes the bug in HC set
+        assert not orig_tcp_props.contained_in(opt_props) or not opt_props.contained_in(orig_tcp_props) or \
+               orig_tcp_props == opt_props
```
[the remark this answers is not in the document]
tanyaveksler: Removed

nca/Parsers/IstioPolicyYamlParser.py
tanyaveksler: Removed the dependency

nca/Parsers/IstioPolicyYamlParser.py
tanyaveksler: Removed the dependency

adisos (changes requested):
additional comments for `ConnectivityProperties.py`

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +571,110 @@ def get_non_tcp_connections():
...
+            conn_cube.unset_dim("dst_peers")
+            if IpBlock.get_all_ips_block() != ip_blocks_mask:
+                src_peers.filter_ipv6_blocks(ip_blocks_mask)
+                dst_peers.filter_ipv6_blocks(ip_blocks_mask)
+            protocols = conn_cube.get_dim("protocols")
+            conn_cube.unset_dim("protocols")
+            if not conn_cube.has_active_dim() and (not protocols or protocols == ignore_protocols):
+                conns = ConnectionSet(True)
+            else:
+                conns = ConnectionSet()
```
adisos, line 617: I do not understand the condition `not protocols` in the `if` statement.
It should be of type `ProtocolSet`. A cube should not have any empty dimension value.
And why do you create `ConnectionSet(True)` if `not protocols` is true?

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+            real_ports = ConnectivityProperties.resolve_named_ports(dst_ports.named_ports, peer, protocols)
+            if real_ports:
+                conn_cube.set_dims({"dst_ports": real_ports, "dst_peers": PeerSet({peer})})
+                conn_properties |= ConnectivityProperties(conn_cube)
+            excluded_real_ports = ConnectivityProperties.resolve_named_ports(dst_ports.excluded_named_ports, peer, protocols)
+            if excluded_real_ports:
+                conn_cube.set_dims({"dst_ports": excluded_real_ports, "dst_peers": PeerSet({peer})})
+                conn_properties -= ConnectivityProperties(conn_cube)
+        return conn_properties
+
```
adisos, line 568: can add unit tests for this named ports resolution during construction?
currently we have unit tests at `testConnectivityPropertiesNamedPorts.py` which only tests the resolving of the original solution.

adisos (commented):
additional comments

nca/Parsers/CalicoPolicyYamlParser.py
```diff
@@ -364,32 +392,61 @@ def _parse_icmp(self, icmp_data, not_icmp_data):
...
         if not_icmp_data is not None:
             self.check_fields_validity(not_icmp_data, 'notICMP', allowed_keys)
-            err = ICMPDataSet.check_code_type_validity(not_icmp_type, not_icmp_code)
+            err = self.check_icmp_code_type_validity(not_icmp_type, not_icmp_code)
             if err:
                 self.syntax_error(err, not_icmp_data)
 
-        res = ICMPDataSet(icmp_data is None and not_icmp_data is None)
+        protocols = ProtocolSet.get_protocol_set_with_single_protocol(protocol)
+        base_peer_set = self.peer_container.get_all_peers_group()
```
adisos, line 405: what if `src_pods` or `dst_pods` contains ip-blocks, which are not part of `base_peer_set`?

nca/Parsers/CalicoPolicyYamlParser.py
```diff
@@ -364,32 +392,61 @@ def _parse_icmp(self, icmp_data, not_icmp_data):
...
+        opt_props = ConnectivityProperties.make_empty_props()
         if icmp_data is not None:
-            res.add_to_set(icmp_type, icmp_code)
+            res = ConnectivityProperties.make_conn_props(conn_cube)
+            if self.optimized_run != 'false':
+                opt_props = ConnectivityProperties.make_conn_props(opt_conn_cube)
             if not_icmp_data is not None:
                 if icmp_type == not_icmp_type and icmp_code == not_icmp_code:
-                    res = ICMPDataSet()
+                    res = ConnectivityProperties.make_empty_props()
```
adisos, line 429: should also set `opt_props` as empty.

nca/Parsers/CalicoPolicyYamlParser.py
```diff
@@ -364,32 +392,61 @@ def _parse_icmp(self, icmp_data, not_icmp_data):
...
                     self.warning('icmp and notICMP are conflicting - no traffic will be matched', not_icmp_data)
                 elif icmp_type == not_icmp_type and icmp_code is None:
-                    tmp = ICMPDataSet()  # this is the only case where it makes sense to combine icmp and notICMP
-                    tmp.add_to_set(not_icmp_type, not_icmp_code)
+                    tmp = ConnectivityProperties.make_conn_props(not_conn_cube)
```
adisos, line 432:
```suggestion
                   # this is the only case where it makes sense to combine icmp and notICMP
                    tmp = ConnectivityProperties.make_conn_props(not_conn_cube)
```

nca/NetworkConfig/NetworkConfig.py
```diff
@@ -265,6 +268,28 @@ def allowed_connections(self, from_peer, to_peer, layer_name=None):
...
+            else:
+                conns_res = self.policies_container.layers[layer_name].allowed_connections_optimized(self.peer_container)
+        else:
+            conns_res = ConnectivityProperties.make_all_props()  # all connections
+            for layer, layer_obj in self.policies_container.layers.items():
+                conns_per_layer = layer_obj.allowed_connections_optimized(self.peer_container)
+                # all allowed connections: intersection of all allowed connections from all layers
+                conns_res &= conns_per_layer
+
+        return conns_res
```
adisos, line 291: should also return allowed captured connections, denied connections, captured flag -- as preparation for next queries to handle.

nca/NetworkConfig/NetworkConfig.py
```diff
@@ -265,6 +268,28 @@ def allowed_connections(self, from_peer, to_peer, layer_name=None):
 
         return allowed_conns_res, captured_flag_res, allowed_captured_conns_res, denied_conns_res
 
+    def allowed_connections_optimized(self, layer_name=None):
```
adisos, line 271: connectivity properties of peers which are HostEndpoints should be split and handled separately, as their connectivity is only determined by Calico layer.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+    def set_dims(self, dims):
+        for dim_name, dim_value in dims.items():
+            self.set_dim(dim_name, dim_value)
+
+    def unset_dim(self, dim_name):
+        assert dim_name in self.dimensions_list
+        self[dim_name] = DimensionsManager().get_dimension_domain_by_name(dim_name)
+
+    def get_dim(self, dim_name):
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+        self.base_peer_set = conn_cube.base_peer_set.copy()
+        if create_empty:
+            return
+        if create_all:  # optimization
+            self.set_all()
+            return
+
+        cube, active_dims, has_empty_dim_value = conn_cube.get_ordered_cube_and_active_dims()
+        if has_empty_dim_value:
+            return
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+from .DimensionsManager import DimensionsManager
+from .PortSet import PortSet
+from .MethodSet import MethodSet
+from .ProtocolSet import ProtocolSet
+from .Peer import PeerSet
+from .ProtocolNameResolver import ProtocolNameResolver
+from .MinDFA import MinDFA
+
+
+class ConnectivityCube(dict):
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+        """
+        :param PeerSet base_peer_set: the set of all possible peers, which will be referenced by the indices
+        in 'src_peers' and 'dst_peers'
+        """
+        super().__init__()
+        self.named_ports = set()  # used only in the original solution
+        self.excluded_named_ports = set()  # used only in the original solution
+        self.base_peer_set = base_peer_set
+        for dim in self.dimensions_list:
+            self[dim] = DimensionsManager().get_dimension_domain_by_name(dim)
```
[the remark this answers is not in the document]
tanyaveksler: Added ConnectivityCube.make_from_dict method

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+         or empty dimension value if the given dimension is not active
+        """
+        res = ConnectivityCube.get_empty_dim(dim_name)
+        if dim_name not in self.active_dimensions:
+            return res
+        for cube in self:
+            conn_cube = self.get_connectivity_cube(cube)
+            values = conn_cube.get_dim(dim_name)
+            if values:
+                res |= values
```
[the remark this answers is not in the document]
tanyaveksler: Fixed.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+        self[dim_name] = DimensionsManager().get_dimension_domain_by_name(dim_name)
+
+    def get_dim(self, dim_name):
+        assert dim_name in self.dimensions_list
+        if dim_name == "src_peers" or dim_name == "dst_peers":
+            if self.is_active_dim(dim_name):
+                # translate CanonicalIntervalSet back to PeerSet
+                return self.base_peer_set.get_peer_set_by_indices(self[dim_name])
+            else:
+                return None
```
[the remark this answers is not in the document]
tanyaveksler: It depends on the usage - for example, in the example code that you brought, `peer_container.get_all_peers_group(True)` will contain `IpBlocks` as well, and they are not a part of base_peer_set.

nca/CoreDS/ConnectionSet.py
tanyaveksler: Right, this was a leftover from some older version. Removed `not protocols` from `if`.

nca/Parsers/IstioPolicyYamlParser.py
```diff
@@ -485,30 +491,43 @@ def parse_ingress_rule(self, rule):
...
+            condition_conns.add_connections('TCP', condition_props)
         if not res_peers:
             self.warning('Rule selects no pods', rule)
-
-        return IstioPolicyRule(res_peers, connections)
+        if not res_peers or not selected_peers:
+            condition_props = ConnectivityProperties.make_empty_props()
+        else:
+            conn_cube = ConnectivityCube(self.peer_container.get_all_peers_group())
+            conn_cube.set_dims({"src_peers": res_peers, "dst_peers": selected_peers})
```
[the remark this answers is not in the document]
tanyaveksler: We only need pods for `base_peer_set `to translate a pod to index. For `IpBlocks `we just calculate the index based on the corresponding ipv4/ipv6 starting index.

nca/Parsers/CalicoPolicyYamlParser.py
tanyaveksler: Reorganized this code, now `opt_props `is initialized as empty.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,603 @@
...
+        """
+        super().__init__()
+        self.named_ports = set()  # used only in the original solution
+        self.excluded_named_ports = set()  # used only in the original solution
+        self.base_peer_set = base_peer_set
+        for dim in self.dimensions_list:
+            self[dim] = DimensionsManager().get_dimension_domain_by_name(dim)
+
+    def copy(self):
+        res = ConnectivityCube(self.base_peer_set.copy())
```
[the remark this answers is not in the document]
tanyaveksler: Done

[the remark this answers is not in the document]
tanyaveksler: This is great, I did not understand why I had those warnings :)

nca/Parsers/K8sPolicyYamlParser.py
```diff
@@ -321,18 +323,40 @@ def parse_ingress_egress_rule(self, rule, peer_array_key):
...
         ports_array = rule.get('ports', [])
         if ports_array:
-            res_ports = ConnectionSet()
+            res_conns = ConnectionSet()
             for port in ports_array:
-                res_ports |= self.parse_port(port)
+                protocol, dest_port_set = self.parse_port(port)
+                if isinstance(protocol, str):
+                    protocol = ProtocolNameResolver.get_protocol_number(protocol)
+                conn_cube = ConnectivityCube(self.peer_container.get_all_peers_group())
```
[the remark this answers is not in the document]
tanyaveksler: After adding `ConnectivityCube.make_from_dict` method it looks much better.

nca/NetworkConfig/NetworkConfig.py
tanyaveksler: Done

nca/NetworkConfig/NetworkConfig.py
tanyaveksler: Changed according to the original implementation.

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -153,12 +166,42 @@ def allowed_connections(self, from_peer, to_peer):
...
+        allowed_ingress_conns |= ConnectivityProperties.make_conn_props(conn_cube)
+        allowed_egress_conns, denied_egress_conns = self._allowed_xgress_conns_optimized(False, peer_container)
+        conn_cube.set_dims({"src_peers": all_ips_peer_set, "dst_peers": all_pods})
+        allowed_egress_conns |= ConnectivityProperties.make_conn_props(conn_cube)
+        res = allowed_ingress_conns & allowed_egress_conns
+        # exclude IpBlock->IpBlock connections
+        conn_cube.set_dims({"src_peers": all_ips_peer_set, "dst_peers": all_ips_peer_set})
+        excluded_conns = ConnectivityProperties.make_conn_props(conn_cube)
+        res -= excluded_conns
+        return res
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -153,12 +166,42 @@ def allowed_connections(self, from_peer, to_peer):
...
+        """
+        Compute per network layer the allowed connections between any relevant peers,
+        considering all layer's policies (and defaults)
+        :param PeerContainer peer_container: the peer container holding the peers
+        :return: all allowed connections
+        :rtype: ConnectivityProperties
+        """
+        all_pods = peer_container.get_all_peers_group()
+        all_ips_peer_set = PeerSet({IpBlock.get_all_ips_block()})
+        conn_cube = ConnectivityCube(peer_container.get_all_peers_group())
```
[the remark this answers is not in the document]
tanyaveksler: explained

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -153,12 +166,42 @@ def allowed_connections(self, from_peer, to_peer):
...
+        conn_cube.set_dims({"src_peers": all_pods, "dst_peers": all_ips_peer_set})
+        allowed_ingress_conns, denied_ingres_conns = self._allowed_xgress_conns_optimized(True, peer_container)
+        allowed_ingress_conns |= ConnectivityProperties.make_conn_props(conn_cube)
+        allowed_egress_conns, denied_egress_conns = self._allowed_xgress_conns_optimized(False, peer_container)
+        conn_cube.set_dims({"src_peers": all_ips_peer_set, "dst_peers": all_pods})
```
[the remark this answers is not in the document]
tanyaveksler: explained

nca/Resources/IstioSidecar.py
```diff
@@ -135,3 +150,30 @@ def _is_sidecar_prior(self, from_peer):
                         self == self.namespace.prior_default_sidecar:
```
shireenf-ibm, line 150: `def _is_sidecar_prior(self, from_peer):` is not used

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -663,59 +671,141 @@ def is_in_subset(self, peer):
...
+
     @staticmethod
     def are_labels_all_included(target_labels, pool_labels):
         for key, val in target_labels.items():
             if pool_labels.get(key) != val:
                 return False
         return True
 
-    def exec(self):
+    def exec(self):  # noqa: C901
```
adisos, line 695: simplify this function, consider splitting to multiple functions

adisos (commented):
consider adding another tests job on the `test-push` workflow that runs all tests with this optimization (maybe twice, one for `debug` and one for `true`).

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +570,109 @@ def get_non_tcp_connections():
...
+        :param ConnectivityProperties conn_props: properties describing allowed connections
+        :param ClusterInfo cluster_info: the cluster info
+        :param PeerContainer peer_container: the peer container
+        :param IpBlock ip_blocks_mask: IpBlock containing all allowed ip values,
+         whereas all other values should be filtered out in the output
+        :param Union[str,None] connectivity_restriction: specify if connectivity is restricted to
+               TCP / non-TCP , or not
+        :return: FWRules map
+        """
+        ignore_protocols = ProtocolSet()
```
adisos, line 590: rename `ignore_protocols` to `relevant_protocols`

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +570,109 @@ def get_non_tcp_connections():
...
+            if IpBlock.get_all_ips_block() != ip_blocks_mask:
+                src_peers.filter_ipv6_blocks(ip_blocks_mask)
+                dst_peers.filter_ipv6_blocks(ip_blocks_mask)
+            protocols = conn_cube["protocols"]
+            conn_cube.unset_dim("protocols")
+            if not conn_cube.has_active_dim() and (protocols.is_whole_range() or protocols == ignore_protocols):
+                conns = ConnectionSet(True)
+            else:
+                conns = ConnectionSet()
+                protocol_names = ProtocolSet.get_protocol_names_from_interval_set(protocols) if protocols else ['TCP']
```
adisos, line 617: how can `protocols` be empty?
why is the default `['TCP']`?

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +570,109 @@ def get_non_tcp_connections():
...
+                for protocol in protocol_names:
+                    if conn_cube.has_active_dim():
+                        conns.add_connections(protocol, ConnectivityProperties.make_conn_props(conn_cube))
+                    else:
+                        if ConnectionSet.protocol_supports_ports(protocol):
+                            conns.add_connections(protocol, ConnectivityProperties.make_all_props())
+                        elif ConnectionSet.protocol_is_icmp(protocol):
+                            conns.add_connections(protocol, ConnectivityProperties.make_all_props())
+                        else:
+                            conns.add_connections(protocol, True)
```
adisos, line 627: the first two branches can be merged to one

nca/FWRules/ConnectivityGraph.py
```diff
@@ -51,6 +68,42 @@ def add_edges(self, connections):
...
+            src_peers.filter_ipv6_blocks(ip_blocks_mask)
+            dst_peers.filter_ipv6_blocks(ip_blocks_mask)
+        protocols = conn_cube["protocols"]
+        conn_cube.unset_dim("protocols")
+
+        if protocols.is_whole_range() and not conn_cube.has_active_dim():
+            conns = ConnectionSet(True)
+        else:
+            conns = ConnectionSet()
+            protocol_names = ProtocolSet.get_protocol_names_from_interval_set(protocols) if protocols else ['TCP']
```
adisos, line 92: same question here (as in the `ConnectionSet`)

nca/Resources/IstioSidecar.py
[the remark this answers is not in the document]
tanyaveksler: Removed

nca/CoreDS/Peer.py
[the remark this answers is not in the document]
tanyaveksler: Removed

nca/CoreDS/Peer.py
```diff
@@ -633,3 +611,123 @@ def by_full_name(elem):
...
+                        if isinstance(cidr.start.address, ipaddress.IPv4Address):
+                            res.add_interval(CanonicalIntervalSet.Interval(self.min_ipv4_index + int(cidr.start),
+                                                                           self.min_ipv4_index + int(cidr.end)))
+                        elif isinstance(cidr.start.address, ipaddress.IPv6Address):
+                            res.add_interval(CanonicalIntervalSet.Interval(self.min_ipv6_index + int(cidr.start),
+                                                                           self.min_ipv6_index + int(cidr.end)))
+                        else:
+                            assert False
+            assert covered_peers == peer_set
+            return res
```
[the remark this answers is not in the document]
tanyaveksler: Optimized

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,640 @@
...
+from .CanonicalHyperCubeSet import CanonicalHyperCubeSet
+from .DimensionsManager import DimensionsManager
+from .PortSet import PortSet
+from .MethodSet import MethodSet
+from .Peer import PeerSet, BasePeerSet
+from .ProtocolNameResolver import ProtocolNameResolver
+from .MinDFA import MinDFA
+
+
+class ConnectivityCube(dict):
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,640 @@
...
+        """
+        return ConnectivityProperties()
+
+    @staticmethod
+    def make_all_props():
+        """
+        Returns all connectivity properties, representing logical True
+        :return: ConnectivityProperties
+        """
+        return ConnectivityProperties(True)
```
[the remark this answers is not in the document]
tanyaveksler: I added the following comment in the `ConnectivityProperties `class:
The `src_peers` and `dst_peers` dimensions are special dimensions,  they do not have constant domain. Their domain
        depends on the current set of peers in the system (as appears in `BasePeerSet` singleton). This set grows when
        adding more configurations. Thus, there is no unique 'all values' representation. In particular, those
        dimensions are never reduced to inactive.
        This might be a problem in comparison and inclusion operators of `ConnectivityProperties`. The possible solution
        may be to keep 'reference full domain value' for these dimensions (as another member in the `BasePeerSet`),
        and to set it to relevant values per query, and to make a special treatment of these dimensions
        in the above operators.

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -101,6 +104,17 @@ def empty_layer_allowed_connections(layer_name, from_peer, to_peer):
         empty_layer_obj = layer_name.create_network_layer([])
         return empty_layer_obj.allowed_connections(from_peer, to_peer)
 
+    @staticmethod
+    def empty_layer_allowed_connections_optimized(peer_container, layer_name):
+        """
+        Get allowed connections between for all relevant peers for an empty layer (no policies).
+        :param PeerContainer peer_container: holds all the peers
+        :param NetworkLayerName layer_name: The empty layer name
+        :rtype: ConnectivityProperties
```
adisos, line 113:
```suggestion
        :rtype: OptimizedPolicyConnections
```

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -191,6 +237,34 @@ def collect_policies_conns(self, from_peer, to_peer, is_ingress,
...
 
+    def collect_policies_conns_optimized(self, is_ingress, captured_func=lambda policy: True):
+        """
+        Collect all connections (between all relevant peers), considering all layer's policies that capture the
+        relevant peers.
+        :param bool is_ingress: indicates whether to return ingress connections or egress connections
+        :param captured_func: callable that returns True if the policy satisfies additional conditions required for
+         considering captured pods instead of applying the default connections.
+        :return: allowed_conns, denied_conns and set of peers to be added to captured peers
+        :rtype: tuple (ConnectivityProperties, ConnectivityProperties, PeerSet)
```
adisos, line 248:
```suggestion
        :rtype: OptimizedPolicyConnections
```

nca/NetworkConfig/NetworkConfigQuery.py
tanyaveksler: Done

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -153,12 +167,44 @@ def allowed_connections(self, from_peer, to_peer):
...
+
     def _allowed_xgress_conns(self, from_peer, to_peer, is_ingress):
         """
         Implemented by derived classes to get allowed and denied ingress/egress connections between from_peer and to_pee
         """
         return NotImplemented
 
+    def _allowed_xgress_conns_optimized(self, is_ingress, peer_container):
+        """
+        Implemented by derived classes to get ingress/egress connections between any relevant peers
```
adisos, line 204:
```suggestion
        Implemented by derived classes to get ingress/egress connections between any relevant peers
        :rtype: OptimizedPolicyConnections
```

nca/CoreDS/ConnectivityProperties.py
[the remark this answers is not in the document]
tanyaveksler: Opened issue https://github.com/IBM/network-config-analyzer/issues/485 about it.

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -212,18 +286,54 @@ def _allowed_xgress_conns(self, from_peer, to_peer, is_ingress):
         return PolicyConnections(captured_res, allowed_conns, denied_conns,
                                  all_allowed_conns=allowed_conns | allowed_non_captured_conns)
 
+    def _allowed_xgress_conns_optimized(self, is_ingress, peer_container):
+        res_conns = self.collect_policies_conns_optimized(is_ingress)
+        # Note: The below computation of non-captured conns cannot be done during the parse stage,
+        # since before computing non-captured conns we should collect all policies conns
+        # compute non-captured connections
```
adisos, line 293:
```suggestion
        # since before computing non-captured conns we should collect all policies conns

        # compute non-captured connections
```

nca/Resources/NetworkPolicy.py
```diff
@@ -272,3 +332,17 @@ class PolicyConnections:
     denied_conns: ConnectionSet = ConnectionSet()  # Connections denied by the policy(ies)
     pass_conns: ConnectionSet = ConnectionSet()  # Connections specified as PASS by the policy(ies)
     all_allowed_conns: ConnectionSet = ConnectionSet()  # all (captured+ non-captured) Connections allowed by the policy(ies)
+
+
+# TODO - making OptimizedPolicyConnections a dataclass does not work
+# (probably because PeerSet and ConnectivityProperties are mutable)
+class OptimizedPolicyConnections:
```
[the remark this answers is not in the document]
tanyaveksler: Dataclasses with mutable members share the same instances of the members. That's not what we want, we want to change the members of `OptimizedPolicyConnections`.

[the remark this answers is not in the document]
tanyaveksler: For example. look in this documentation:
https://docs.python.org/3/library/dataclasses.html#default-factory-functions

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -721,21 +764,105 @@ def exec(self):
...
+                self.get_connectivity_output_split_by_tcp(connections, peers, peers_to_compare)
         else:
-            output_res = self.get_connectivity_output_full(connections, peers, peers_to_compare)
+            output_res, fw_rules = self.get_connectivity_output_full(connections, peers, peers_to_compare)
+        return output_res, fw_rules, fw_rules_tcp, fw_rules_non_tcp
 
-        res = QueryAnswer(True)
-        if self.output_config.outputFormat in ['json', 'yaml']:
-            res.output_explanation = [ComputedExplanation(dict_explanation=output_res)]
+    def compute_connectivity_output_optimized(self):
```
adisos, line 776: add rtype

nca/CoreDS/ConnectionSet.py
```diff
@@ -538,10 +542,26 @@ def print_diff(self, other, self_name, other_name):
...
 
+    def convert_to_connectivity_properties(self, peer_container):
+        if self.allow_all:
+            return ConnectivityProperties.make_all_props()
+
+        res = ConnectivityProperties.make_empty_props()
+        for protocol, properties in self.allowed_protocols.items():
+            protocols = ProtocolSet.get_protocol_set_with_single_protocol(protocol)
+            conn_cube = ConnectivityCube.make_from_dict({"protocols": protocols})
+            this_prop = ConnectivityProperties.make_conn_props(conn_cube)
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectionSet.py
```diff
@@ -550,3 +570,109 @@ def get_non_tcp_connections():
...
+                # TODO try to split the element below by labels
+                res.append(FWRule.PeerSetElement(ns_peers & peer_set_copy))
+            peer_set_copy -= ns_peers
+        if ns_set:
+            res.append(FWRule.FWRuleElement(ns_set))
+
+        return res
+
+    @staticmethod
+    def fw_rules_to_conn_props(fw_rules, peer_container):
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectionSet.py
tanyaveksler: Done

nca/CoreDS/ConnectionSet.py
tanyaveksler: Changed. Added an assertion about it.

nca/CoreDS/ConnectionSet.py
tanyaveksler: Done

nca/FWRules/ConnectivityGraph.py
```diff
@@ -26,13 +29,27 @@ def __init__(self, all_peers, allowed_labels, output_config):
...
+        :param Peer peer: the peer object
+        :return: tuple(str, bool)
+        str: the peer name
+        bool: flag to indicate if peer is ip-block (True) or not (False)
+        """
+        if isinstance(peer, IpBlock):
+            return peer.get_ip_range_or_cidr_str(), True
+        if self.output_config.outputEndpoints == 'deployments' and isinstance(peer, Pod):
+            return peer.workload_name, False
+        return str(peer), False
```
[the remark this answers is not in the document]
tanyaveksler: Removed

nca/FWRules/ConnectivityGraph.py
tanyaveksler: Added assertion (as in the `ConnectionSet`)

nca/FWRules/ConnectivityGraph.py
```diff
@@ -51,6 +68,42 @@ def add_edges(self, connections):
         """
         self.connections_to_peers.update(connections)
 
+    def add_edges_from_cube_dict(self, conn_cube, ip_blocks_mask):
```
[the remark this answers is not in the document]
tanyaveksler: Moved the common functionality to a separate function `ConnectivityProperties.get_connection_set_and_peers_from_cube()`

nca/FWRules/ConnectivityGraph.py
```diff
@@ -51,6 +68,42 @@ def add_edges(self, connections):
...
+        if IpBlock.get_all_ips_block() != ip_blocks_mask:
+            src_peers.filter_ipv6_blocks(ip_blocks_mask)
+            dst_peers.filter_ipv6_blocks(ip_blocks_mask)
+        protocols = conn_cube["protocols"]
+        conn_cube.unset_dim("protocols")
+
+        if protocols.is_whole_range() and not conn_cube.has_active_dim():
+            conns = ConnectionSet(True)
+        else:
+            conns = ConnectionSet()
```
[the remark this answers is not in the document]
tanyaveksler: Currently `ConnectivityGraph` is built from `ConnectionSet`. This will be changed as a part of removing `ConnectionSet`.

nca/FWRules/MinimizeFWRules.py
```diff
@@ -667,6 +667,10 @@ def __init__(self, fw_rules_map, cluster_info, output_config, results_map):
         self.output_config = output_config
         self.results_map = results_map
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
tanyaveksler: Removed. Not needed anymore.

nca/FWRules/ClusterInfo.py
```diff
@@ -52,6 +52,12 @@ def __init__(self, all_peers, allowed_labels):
         # labels values for a combination of multiple labels, in a fw-rule
         self.add_update_pods_labels_map_with_required_conjunction_labels()
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
tanyaveksler: Removed. Not needed anymore.

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -235,6 +345,36 @@ def captured_cond_func(policy):
         return PolicyConnections(captured_res, allowed_conns, denied_conns,
                                  all_allowed_conns=allowed_conns | allowed_non_captured_conns)
 
+    def _allowed_xgress_conns_optimized(self, is_ingress, peer_container):
+        res_conns = self.collect_policies_conns_optimized(is_ingress, IstioNetworkLayer.captured_cond_func)
+        all_peers_and_ips = peer_container.get_all_peers_group(True)
+        dns_entries = peer_container.get_all_dns_entries()
+        # for istio initialize non-captured conns with non-TCP connections
+        conn_cube = ConnectivityCube.make_from_dict({"src_peers": all_peers_and_ips, "dst_peers": all_peers_and_ips,
+                                                     "protocols": ProtocolSet.get_non_tcp_protocols()})
```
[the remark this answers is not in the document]
tanyaveksler: It's just a shortcut to contain connections from peers to ip-blocks and vice-versa. This is assuming that at the end the ip-block to ip-block connections will be filtered out.

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -235,6 +345,36 @@ def captured_cond_func(policy):
...
+        res_conns.all_allowed_conns |= res_conns.allowed_conns | ConnectivityProperties.make_conn_props(conn_cube)
+        non_captured_peers = all_peers_and_ips - res_conns.captured
```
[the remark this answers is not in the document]
tanyaveksler: Fixed

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -235,6 +345,36 @@ def captured_cond_func(policy):
...
+        if non_captured_peers:
+            protocols = ProtocolSet.get_protocol_set_with_single_protocol('TCP')
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/NetworkConfig/NetworkConfig.py
```diff
@@ -265,6 +269,49 @@ def allowed_connections(self, from_peer, to_peer, layer_name=None):
...
+        if layer_name is not None:
+            if layer_name not in self.policies_container.layers:
+                return self.policies_container.layers.empty_layer_allowed_connections_optimized(self.peer_container,
+                                                                                                layer_name)
+            return self.policies_container.layers[layer_name].allowed_connections_optimized(self.peer_container)
+
+        all_peers = self.peer_container.get_all_peers_group()
+        host_eps = Peer.PeerSet(set([peer for peer in all_peers if isinstance(peer, Peer.HostEP)]))
+        src_peers_conn_cube = ConnectivityCube.make_from_dict({"src_peers": host_eps})
+        dst_peers_conn_cube = ConnectivityCube.make_from_dict({"dst_peers": host_eps})
```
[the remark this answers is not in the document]
tanyaveksler: These variables are no more created

nca/NetworkConfig/NetworkConfig.py
```diff
@@ -265,6 +269,49 @@ def allowed_connections(self, from_peer, to_peer, layer_name=None):
...
+        # all possible connections involving hostEndpoints
+        conn_hep = ConnectivityProperties.make_conn_props(src_peers_conn_cube) | \
+            ConnectivityProperties.make_conn_props(dst_peers_conn_cube)
+        conns_res = OptimizedPolicyConnections()
+        conns_res.all_allowed_conns = ConnectivityProperties.make_all_props()
```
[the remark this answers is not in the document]
tanyaveksler: Done. Also in other places.

nca/Resources/NetworkPolicy.py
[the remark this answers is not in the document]
adisos: why not using `default_factory`?

nca/NetworkConfig/NetworkLayer.py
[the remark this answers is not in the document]
adisos: please add a comment explaining this in the code.

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -746,13 +873,30 @@ def get_connectivity_output_full(self, connections, peers, peers_to_compare):
...
+        return formatted_rules, fw_rules
+
+    def get_props_output_full(self, props, peers_to_compare, ip_blocks_mask):
+        """
+        get the connectivity map output considering all connections in the output
+        :param ConnectivityProperties props: properties describing allowed connections
+        :param PeerSet peers_to_compare: the peers to consider for dot/fw-rules output
+        :param IpBlock ip_blocks_mask:  IpBlock containing all allowed ip values,
+         whereas all other values should be filtered out in the output
+        :rtype Union[str,dict]
```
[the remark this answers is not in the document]
tanyaveksler: Done.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,472 @@
...
+                continue
+            if real_port[1] not in protocols:
+                print(f'Warning: Illegal protocol {real_port[1]} in the named port {named_port} '
+                      f'of the pod {peer}. Ignoring the pod')
+                continue
+            real_ports.add_port(real_port[0])
+        return real_ports
+
+    @staticmethod
+    def make_conn_props(conn_cube):
```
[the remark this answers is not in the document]
tanyaveksler: Renamed to `_make_conn_props_no_named_ports_resolution`.

nca/NetworkConfig/NetworkConfig.py
```diff
@@ -287,7 +287,10 @@ def allowed_connections_optimized(self, layer_name=None):
         conn_hep = ConnectivityProperties.make_conn_props_from_dict({"src_peers": host_eps}) | \
             ConnectivityProperties.make_conn_props_from_dict({"dst_peers": host_eps})
         conns_res = OptimizedPolicyConnections()
-        conns_res.all_allowed_conns = ConnectivityProperties.make_all_props()
+        all_peers_and_ips_and_dns = self.peer_container.get_all_peers_group(True, True, True)
+        conns_res.all_allowed_conns = \
+            ConnectivityProperties.make_conn_props_from_dict({"src_peers": all_peers_and_ips_and_dns,
+                                                              "dst_peers": all_peers_and_ips_and_dns})
```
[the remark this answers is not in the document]
tanyaveksler: Done

adisos (commented):
few more comments remaining:

- new tests workflow to run with debug flag and check correctness of optimized implementation

https://github.com/IBM/network-config-analyzer/pull/447#discussion_r1176041953

https://github.com/IBM/network-config-analyzer/pull/447#discussion_r1174849823

nca/NetworkConfig/NetworkLayer.py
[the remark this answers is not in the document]
tanyaveksler: Done. Also in other similar place.

nca/Resources/NetworkPolicy.py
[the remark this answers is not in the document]
tanyaveksler: I tried using `default_factory`, it still did not work. I think the problem is that I am using the data members of `OptimizedPolicyConnections` class not only for keeping data but also for updating data, for example in `IstioNetworkLayer._allowed_xgress_conns_optimized`, like this:
`res_conns.all_allowed_conns |= all_nc_conns - res_conns.denied_conns`.
This somehow changes the default values of the dataclass.
A possible solution, is we insist on using dataclasses, is using another variables for intermediate calculations, but I don't this it's worthwhile.

.github/workflows/test-push.yml
```diff
@@ -87,6 +87,27 @@ jobs:
...
+      - name: install helm
+        run: |
+          curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
+          chmod 700 get_helm.sh
+          sudo ./get_helm.sh
+      - name: Run k8s tests
+        env:
+          GHE_TOKEN: ${{ github.token }}
+          PYTHONPATH: .
+        run: python tests/run_all_tests.py --type=general --category=k8s --hc_opt=debug --check_run_time | tee tests/k8s_cmp_log.txt ; test ${PIPESTATUS[0]} -eq 0
```
adisos, line 104: no need to run with `--check_run_time`

nca/SchemeRunner.py
```diff
@@ -190,6 +190,11 @@ def run_queries(self, query_array):
             not_executed = 0
             self.check_fields_validity(query, 'query', allowed_elements)
             query_name = query['name']
+            if self.optimized_run == 'debug':
+                #optimization - currently only connectivityMap query has optimized implementation and can be compared
+                if not query.get('connectivityMap'):
+                    print(f'Skipping query {query_name} since it does not have optimized implementation yet')
+                    continue
```
adisos: add a `TODO` comment to remove this after supporting additional queries.
Also, create an issue for supporting additional queries, and a link to this comment in the issue.

tanyaveksler: Done

.github/workflows/test-push.yml
tanyaveksler: Removed `--check_run_time`

nca/FWRules/ConnectivityGraph.py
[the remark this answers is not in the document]
adisos: This is currently a static method of `ConnectionSet`. Shouldn't be moved as a method of `ConnectivityProperties` ?

nca/FWRules/ConnectivityGraph.py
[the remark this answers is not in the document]
adisos: Create another issue to consider the removal of `ConnectionSet` after supporting all queries with optimized solution, as preparation for removing original solution.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,483 @@
...
+        else:
+            res = DimensionsManager().get_empty_dimension_by_name(dim_name)
+        for cube in self:
+            conn_cube = self.get_connectivity_cube(cube)
+            values = conn_cube[dim_name]
+            if values and res:
+                res |= values
+            elif values:
+                res = values
+        return res
```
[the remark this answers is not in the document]
tanyaveksler: Yes, good point. Changed to assertion.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,483 @@
...
+                continue
+            real_ports.add_port(real_port[0])
+        return real_ports
+
+    @staticmethod
+    def make_conn_props(conn_cube):
+        """
+        This will create connectivity properties made of the given connectivity cube.
+        This includes tcp properties, non-tcp properties, icmp data properties.
+        If possible (i.e., in original solution, when dst_peers are supported), the named ports will be resolved.
```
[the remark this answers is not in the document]
tanyaveksler: Right, this is a typo - I meant in the optimized solution. Fixed.

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,483 @@
...
+            conn_cube = self.get_connectivity_cube(cube)
+            values = conn_cube[dim_name]
+            if values and res:
+                res |= values
+            elif values:
+                res = values
+        return res
+
+    @staticmethod
+    def resolve_named_ports(named_ports, peer, protocols):
```
adisos, line 369: rename method to `_resolve_named_ports`

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -721,46 +772,146 @@ def exec(self):
...
+        return output_res, fw_rules, fw_rules_tcp, fw_rules_non_tcp
 
-        res = QueryAnswer(True)
-        if self.output_config.outputFormat in ['json', 'yaml']:
-            res.output_explanation = [ComputedExplanation(dict_explanation=output_res)]
+    def compute_connectivity_output_optimized(self):
+        """
+        Compute connectivity output with optimized implementation.
+        :return: a tuple of output result (in a required format), FwRules, tcp FWRules and non-tcp FWRules.
+        :rtype: ([Union[str, dict], MinimizeFWRules, MinimizeFWRules], MinimizeFWRules)
```
adisos, line 788:
```suggestion
        :rtype: ([Union[str, dict], MinimizeFWRules, MinimizeFWRules, MinimizeFWRules)
```

nca/CoreDS/ConnectivityProperties.py
```diff
@@ -0,0 +1,483 @@
...
+                real_port_number = real_port[0]
+                rectangle = [my_excluded_named_ports[port],
+                             CanonicalIntervalSet.get_interval_set(real_port_number, real_port_number)]
+                self.add_hole(rectangle, active_dims)
+
+    def copy(self):
+        """
+        :rtype: ConnectivityProperties
+        """
+        res = ConnectivityProperties._make_conn_props_no_named_ports_resolution(ConnectivityCube())
```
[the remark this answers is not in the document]
tanyaveksler: Done

nca/CoreDS/ConnectivityProperties.py
tanyaveksler: Done

nca/NetworkConfig/NetworkConfig.py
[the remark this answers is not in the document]
adisos: There are more calls to `ConnectivityProperties.make_all_props()` in `IstioPolicyYamlParser.py`.
Should they also be replaced with `get_all_conns_props_per_config_peers` ?

adisos (approved):

PR #448, opened 2023-02-27
supporting txt_no_fw_rules for configs with istio layer

shireenf-ibm:


adisos (commented):
please also update the message in `MinimizeFWRules`, currently it is ` unexpected outputFormat in output configuration value [should be txt/yaml/csv]` and it does not take into account other supported formats.

nca/FWRules/MinimizeFWRules.py
```diff
@@ -668,7 +668,7 @@ def get_fw_rules_in_required_format(self, add_txt_header=True, add_csv_header=Tr
             query_name += ', config: ' + self.output_config.configName
         output_format = self.output_config.outputFormat
         if output_format not in FWRule.supported_formats:
-            print(f'error: unexpected outputFormat in output configuration value [should be txt/yaml/csv],  '
+            print(f'error: unexpected outputFormat in output configuration value [should be txt/yaml/csv/md/json],  '
```
adisos, line 671: missing the new `txt_no_fw_rules` format here.
is it possible to get the supported formats from where they are defined instead of duplicating them here?
This duplication is error-prone.

shireenf-ibm: I didn't add the `txt_no_fw_rules` intentionally, this `FWRules def` is not reachable for` txt_no_fw_rules, dot, jpg `formats (these formats don't use the FWrules/ minimize them)

adisos: But if I have a typo in the specified output format , nca will reach this error, even though it should not necessarily be a fw-rules format. can it be checked/caught earlier, so the error message displays all possible output formats?

shireenf-ibm: I see, you are right 
only when returning from the "query.execute" will get the message of `<format with typo> output format is not supported for this query`
I may list all formats again here (even if the format does not use the fw_rules)
or might call this def only if the format supports FW-rule (add if statements in the relevant queries calling this def)

adisos: I think it may be preferred to check even before the query is executed

shireenf-ibm: previously, we executed the query regardless of the `outputFormat `(an error was raised only if it gets to `MinimizeFWRules`)
In last commit if the format is not supported , the query will not be executed, and return with the not_executed flag so the relevant message of `outputConfiguration`:  `<format with typo> output format is not supported for this query` will be printed

adisos (approved):

nca/FWRules/ConnectivityGraph.py
```diff
@@ -154,10 +154,10 @@ def _creates_cliqued_graph(directed_edges):
 
         return directed_edges, not_directed_edges, cliques_nodes
 
-    def get_connections_without_fw_rules_txt_format(self, connectivity_restriction=None):
+    def get_connections_without_fw_rules_txt_format(self, connectivity_msg=None):
         """
-        :param Union[str,None] connectivity_restriction: specify if connectivity is restricted to
-               TCP / non-TCP , or not
+        :param Union[str,None] connectivity_msg: a msg header describing either the type of connectivity (TCP/non-TCP)
+            or the type of connectivity changes
```
adisos, line 160:
```suggestion
        :param Union[str,None] connectivity_msg: a msg header describing either the type of connectivity (TCP/non-TCP)
            for connectivity-map output with connectivity restriction,  or the type of connectivity changes for semantic-diff query output
```

adisos (changes requested):
just a small comment for more details in documentation

adisos (approved):

PR #467, opened 2023-03-29
Service entry initial support

shireenf-ibm:
This PR contains examples only with queries which work on one network-config only (e.g. ConnectivityMap, Sanity ...).
It doesn't contain examples of queries with two network-configs, since currently, the prior_sidecar of each peer is an attribute of the peer itself and not of the config, and this may cause a bug computing the egress connections of a single peer. 
a case that exemplifies the bug:
- suppose we have:
    -  a pod called `ratings`
    - first network config with a sidecar that selects `ratings` and restricts its egress connections to be enabled only to another pod called `reviews`
     - second network config with a sidecar that selects `ratings` pod and enables egress connections from it to any endpoint
 
 so by computing allowed connections of the  `ratings` pod on the two configurations , we expect to see differences.
 but since the sidecar is an attribute of the pod, and when executing the computations of a query with two-configs we bring the common peers only once from one config (the first), we'll see that the connections are the same.
 
 In the next PR, we'll change relating the sidecar with the peer itself, so the code of the queries executing will not miss the different allowed connections between two-configs


nca/NetworkConfig/PeerContainer.py
```diff
@@ -216,26 +250,35 @@ def get_pods_with_service_name_containing_given_string(self, name_substring):
                 res |= val.target_pods
         return res
 
-    def get_all_services_target_pods(self):
+    def get_all_services_target_peers(self):
         """
-        Returns all pods that belong to services
+        Returns all pods that belong to services and all DNSEntry peers (if existed) as this type of peers
+        is produced by ServiceEntry objects
```
adisos, line 256:
```suggestion
        Returns all pods that belong to services and all DNSEntry peers (if exist) as this type of peers
        is produced by ServiceEntry objects
```

nca/NetworkConfig/PeerContainer.py
```diff
@@ -216,26 +250,35 @@ def get_pods_with_service_name_containing_given_string(self, name_substring):
...
         return res
 
-    def get_services_target_pods_in_namespace(self, namespace):
+    def get_services_target_peers_in_namespace(self, namespace):
         """
-        Returns all pods that belong to services in the given namespace
-        :param K8sNamespace namespace:   namespace object
+        Returns all peers that belong to services in/ exported to the given namespace
+        (i.e. pods that belong to k8s services in the given namespace,
+        or DNSEntry peers)
```
adisos, line 271:
```suggestion
        or DNSEntry peers)
        currently assuming DNSEntry peers are exported to all namespaces
```

adisos (commented):
a few initial comments

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -142,6 +142,25 @@ def execute(self, cmd_line_flag):
...
+        :param Peer peer2: the dst peer
+        :rtype: bool
+        """
+        if isinstance(peer1, DNSEntry):  # connections from DNSEntry are not relevant
+            return False
+        if isinstance(peer1, IpBlock) and isinstance(peer2, (IpBlock, DNSEntry)):
+            return False  # connectivity between external peers is not relevant either
+        if not self.config.policies_container.layers.does_contain_layer(NetworkLayerName.Istio) \
+                and isinstance(peer2, DNSEntry):
+            return False  # connectivity to DNSEntry peers is only relevant if istio layer exists
```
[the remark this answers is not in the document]
shireenf-ibm: I think this will be easiest to be done in the next PRs (3rd pr)
in this basic PR, we parse serviceentry only for peers , we don't create policy elements from a service entry.
whith supporting the "exportTo" and "ports" fields , I am planing to add a policy elements when parsing the serviceEntry, then these policies will be declared to be in istio layer

nca/NetworkConfig/PeerContainer.py
```diff
@@ -204,6 +204,40 @@ def get_target_pods_with_service_name(self, service_name):
...
+        translates the host name (dns) to a pattern that may be used with re library methods
+         - "*" at the beginning of the host name will be replaced with an FQDN pattern
+         - the "." in the host name may not be replaced by any other character
+         :param str host_name: the host name
+         :return: the re-pattern of the host name
+         :rtype: str
+        """
+        if host_name.startswith('*'):
+            name_suffix = re.escape(host_name[1:])
+            return r"(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?" + name_suffix
```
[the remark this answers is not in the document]
shireenf-ibm: I may add it as a static variable of `DNSEntry `class and use it in both places

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -142,6 +142,25 @@ def execute(self, cmd_line_flag):
     def exec(self):
         raise NotImplementedError
 
+    # this def contains conditions that should be checked every time before computing
+    # allowed connections of two peers, so added it here to avoid duplications in the queries code
+    def determine_whether_to_compute_allowed_conns_for_peer_types(self, peer1, peer2):
```
adisos, line 147: change so it is used only where relevant

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -861,6 +881,20 @@ def split_to_tcp_and_non_tcp_conns(conns):
...
+        :param list peers_list: list of peers pairs
+        :rtype: list
+        """
+        peers_list_wo_dns = []
+        for peers_pair in peers_list:
+            # a DNSEnty peer may only be in the dest (peers_pair[1])
+            if not isinstance(peers_pair[1], DNSEntry):
+                peers_list_wo_dns.append(peers_pair)
+        return peers_list_wo_dns
+
```
adisos, line 897: consider instead return from `allowed_conns` only TCP for a connection to `serviceEntry`

nca/NetworkConfig/PeerContainer.py
```diff
@@ -204,6 +204,40 @@ def get_target_pods_with_service_name(self, service_name):
                 res |= service.target_pods
         return res
 
+    @staticmethod
+    def _compute_re_pattern_from_host_name(host_name):
```
adisos, line 208: consider move to more relevant class

nca/NetworkConfig/PeerContainer.py
```diff
@@ -204,6 +204,40 @@ def get_target_pods_with_service_name(self, service_name):
...
+        returns all DNSentry peers which are equal/contained in the host_name
+        :param str host_name: the host name as it appears in the policy's hosts list
+        :rtype: PeerSet
+        """
+        res = PeerSet()
+        # re.fullmatch behavior in accordance with live-cluster behaviour.
+        # only sidecar host should contain the service-entry host.
+        # the opposite containment direction is not considered for connections
+        # (detailed examples in:
+        # tests/istio_testcases/example_policies/bookinfo-demo/sidecar_examples/bookinfo-test-sidecar-connectivity-scheme.yaml)
```
adisos, line 234: can add query name relevant to this comment?

nca/NetworkConfig/NetworkConfigQuery.py
shireenf-ibm: done

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -229,6 +229,9 @@ def captured_cond_func(policy):
         if not captured_res:  # no allow policies for target
             # add connections allowed by default that are not captured
             allowed_non_captured_conns |= (ConnectionSet(True) - denied_conns)
+            # exception: update allowed non-captured conns to DNSEntry dst with TCP only
+            if isinstance(to_peer, DNSEntry):
+                allowed_non_captured_conns = ConnectionSet.get_all_tcp_connections()
```
adisos, line 234: why is it only in non-captured conns? isn't it also relevant for connections returned from `IstioSidecar.allowed_connections()`?

shireenf-ibm: it is , didn't see a need to currently change since sidecar is an istio policy (then considered automatically as TCP only )

adisos: where is it considered as TCP only? It seems that `IstioSidecar.allowed_connections()` returns `All connections`, and that `IstioNetworkLayer._allowed_xgress_conns()` returns `All connections` for connections captured to `DNSEntry` and `TCP` for connections not captured to `DNSEntry`?

shireenf-ibm: oh, I assumed by splitting to tcp and non-tcp in the ConnectivityMapQuery, all istio's layer conns are TCP 
committed a fix

adisos: still not sure I understand, how did it work (passed the tests) before this fix? 
and why do you restrict sidecar connections to `TCP` only for `DNSentry`?

shireenf-ibm: I agree, I assumed that it worked previously because by splitting connections to TCP and non-TCP in ConnectivityMapQuery , all istio's policies conns (not only from sidecar) were considered TCP. (the splitting occurs only if there is an istio layer, so I thought that all captured conns in that case were considered TCP and what is not captured will be added to the non-TCP section).

I restricted only DNSEntry to TCP, since sidecar itself does not restrict conns 
but serviceEntry does, since it supports only TCP protocols ( when we add the support of ports field this will be more clear) 
so we don't want to see "DNS entries" in the non-TCP section

shireenf-ibm: As I understand from `def split_to_tcp_and_non_tcp_conns(conns):`  in `ConnectivityMapQuery`
it adds as a default `All Connections` for all protocols **other than TCP for** : for all the peers pairs which were captured in the `istio policies` (the captured ones are presented in the TCP section)
and by default all connections are added for both sections (TCP and non-TCP) ,for the non-captured peers at all. 

Note: the `DNSEntry peers` are exceptional, as we don't want to see them in the `non-TCP` section at all (these peers are relevant only for `istio` layer and conns to them may be enabled only with `TCP`, not enabled for other protocols at all ), then we have to make sure they always return as default only TCP connections (and not all conns for all protocols)

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -431,8 +450,6 @@ def other_policy_containing_deny(self, self_policy, config_with_self_policy, lay
                                                       config_with_other_policy).disjoint_referenced_ip_blocks()
             for pod1 in pods_to_compare:
                 for pod2 in pods_to_compare:
-                    if isinstance(pod1, IpBlock) and isinstance(pod2, IpBlock):
-                        continue
```
adisos, line 435: avoid this deletion

adisos (approved):

PR #477, opened 2023-04-13
Explainability with hcs

shmfr:


adisos (commented):
Initial comments:
- [x] fix the `Default` policy issue - to be added only when relevant in the output
- [ ] add documentation, explain core data structures and computation logic

nca/Utils/NcaLogger.py
tanyaveksler, line 1: Update param 'level' description to add 'E' value

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,419 @@
...
+            conn_cube = self.all_conns.get_connectivity_cube(cube)
+            src_peers = conn_cube['src_peers']
+            dst_peers = conn_cube['dst_peers']
+            for peer in src_peers:
+                src_peers_names.append(peer.full_name())
+            for peer in dst_peers:
+                dst_peers_names.append(peer.full_name())
+            if src in src_peers_names and dst in dst_peers_names:
+                return True
+        return False
```
tanyaveksler, line 210: `return True if self.all_conns & ConnectivityProperties.make_conn_props_from_dict({"src_peers": PeerSet({src}), "dst_peers": PeerSet({dst})) else False`
```suggestion
        for cube in self.all_conns:
            src_peers_names = []
            dst_peers_names = []
            conn_cube = self.all_conns.get_connectivity_cube(cube)
            src_peers = conn_cube['src_peers']
            dst_peers = conn_cube['dst_peers']
            for peer in src_peers:
                src_peers_names.append(peer.full_name())
            for peer in dst_peers:
                dst_peers_names.append(peer.full_name())
            if src in src_peers_names and dst in dst_peers_names:
                return True
        return False
```

tanyaveksler, line 210:
```suggestion
return True if self.all_conns & ConnectivityProperties.make_conn_props_from_dict({"src_peers": PeerSet({src}), "dst_peers": PeerSet({dst})) else False```

File: nca/Utils/ExplTracker.py [hunk 2]
```diff
@@ -0,0 +1,419 @@
...
+        :return: str: the explanation out string
+        """
+        if len(nodes) < 1:
+            return ''
+        elif len(nodes) > 2:
+            NcaLogger().log_message(f'Explainability error: only 1 or 2 nodes are allowed for explainability query,'
+                                    f' found {len(nodes)} ', level='E')
+            return ''
+
+        if nodes[0] == 'ALL':
```
tanyaveksler, line 372: The third mode? Should be added to the description

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -35,12 +36,14 @@ def update(self, iteration_results):
         self.query_iterations_output.append(iteration_results[1])
         self.num_not_executed += iteration_results[2]
 
-    def compute_final_results(self, output_format):
+    def compute_final_results(self, output_format, expl_nodes=None):
         """
         extracts the final query results from self variables
         from self.query_iterations_output computes the final str output of the query,
         other results returned as is from query_result.
+        :param expl_nodes: nodes to explain the connectivity between them
```
tanyaveksler, line 44:
```suggestion
```

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -169,7 +176,10 @@ def _run_query_for_each_config(self):
         query_result = QueryResult()
         for config in self.configs_array:
             query_result.update(self._execute_one_config_query(self.query_name, self._get_config(config)))
-        return query_result.compute_final_results(self.output_configuration.outputFormat)
+        expl = ''
+        if self.output_configuration.explain:
+            expl = self.output_configuration.explain
+        return query_result.compute_final_results(self.output_configuration.outputFormat, expl)
```
tanyaveksler, line 182:
```suggestion
        return query_result.compute_final_results(self.output_configuration.outputFormat, self.output_configuration.explain)
```

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -252,6 +253,17 @@ def collect_policies_conns_optimized(self, is_ingress, captured_func=lambda poli
         """
         res_conns = OptimizedPolicyConnections()
         for policy in self.policies_list:
+            # Track the peers that were affected by this policy
+            if ExplTracker().is_active():
+                for peer in policy.selected_peers:
+                    src_peers, _ = ExplTracker().extract_peers(policy.optimized_allow_ingress_props)
+                    _, dst_peers = ExplTracker().extract_peers(policy.optimized_allow_egress_props)
```
tanyaveksler, line 260: What about `policy.optimized_deny_ingress_props` and `policy.optimized_deny_egress_props`? They may also explain a lack of connections between certain peers

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -813,11 +814,15 @@ def compute_connectivity_output_optimized(self):
                            ConnectivityProperties.make_conn_props_from_dict({"dst_peers": subset_peers})
             all_conns_opt &= subset_conns
         all_conns_opt = self.filter_conns_by_peer_types(all_conns_opt, opt_peers_to_compare)
+        expl_conns = all_conns_opt
         if self.config.policies_container.layers.does_contain_layer(NetworkLayerName.Istio):
             output_res, opt_fw_rules_tcp, opt_fw_rules_non_tcp = \
                 self.get_props_output_split_by_tcp(all_conns_opt, opt_peers_to_compare)
+            expl_conns, _ = self.convert_props_to_split_by_tcp(all_conns_opt)
```
tanyaveksler, line 821: What about non-tcp conns?

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -968,7 +975,13 @@ def get_props_output_split_by_tcp(self, props, peers_to_compare):
             # concatenate the two graphs into one dot file
             res_str = dot_tcp + dot_non_tcp
             return res_str, None, None
-        # TODO - handle 'txt_no_fw_rules' output format
+        if self.output_config.outputFormat in ['txt_no_fw_rules']:
+            txt_no_fw_rules_tcp = self.txt_no_fw_rules_format_from_props(props_tcp, peers_to_compare, connectivity_tcp_str)
+            txt_no_fw_rules_non_tcp = self.txt_no_fw_rules_format_from_props(props_non_tcp, peers_to_compare,
+                                                                             connectivity_non_tcp_str)
+            # concatenate the two graphs into one dot file
```
tanyaveksler, line 982:
```suggestion
```

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1035,6 +1048,20 @@ def dot_format_from_props(self, props, peers, connectivity_restriction=None):
...
         return conn_graph.get_connectivity_dot_format_str(connectivity_restriction)
 
+    def txt_no_fw_rules_format_from_props(self, props, peers, connectivity_restriction=None):
+        """
+        :param ConnectivityProperties props: properties describing allowed connections
+        :param PeerSet peers: the peers to consider for dot output
+        :param Union[str,None] connectivity_restriction: specify if connectivity is restricted to
+               TCP / non-TCP , or not
+        :rtype str
+        :return the connectivity map in dot-format, considering connectivity_restriction if required
```
tanyaveksler, line 1058: It's not dot format

nca/Utils/NcaLogger.py
shmfr: Done.

nca/Utils/ExplTracker.py
shmfr: Done (added functionality to extract Peer by name since the suggested solution needs the Peer objects and not peer names).

nca/Utils/ExplTracker.py
shmfr: Done (added the 'ALL' option to the description)

nca/NetworkConfig/NetworkConfigQueryRunner.py
shmfr: Done.

nca/NetworkConfig/NetworkConfigQueryRunner.py
shmfr: Done.

nca/NetworkConfig/NetworkLayer.py
shmfr: In the case of lack of connectivity we show all the policies that affects the queried peers, since also a typo in a 'non deny' policy may cause the lack of connectivity.

nca/NetworkConfig/NetworkConfigQuery.py
shmfr: At this phase, when Istio is present, we 'explain' only the tcp connections.

nca/NetworkConfig/NetworkConfigQuery.py
shmfr: Done.

nca/NetworkConfig/NetworkConfigQuery.py
shmfr: Right :)   Done.

adisos (commented):
Initial comments:
- [ ] add some documentation / update readme files about this new option and how it can be used.
- [ ] open issue with follow-up items for this feature, such as more accurate explanation of denied connection when possible (rather than all policies), explain non-tcp conns even if istio is present, etc.

nca/NetworkConfig/NetworkConfigQueryRunner.py
```diff
@@ -54,7 +56,11 @@ def compute_final_results(self, output_format):
                                sort_keys=False)
         else:
             output = '\n'.join(self.query_iterations_output)
-        return self.numerical_result, output, self.num_not_executed
+        if ExplTracker().is_active() and expl_nodes and (output_format == 'txt' or output_format == 'txt_no_fw_rules'):
+            expl_out = ExplTracker().explain(expl_nodes.split(','))
+        else:
+            expl_out = ''
+        return self.numerical_result, output+expl_out, self.num_not_executed
```
adisos, line 63: why do we need both active/inactive mode for `ExplTracker` and `expl_nodes` from output configuration to check before adding explanation in the output? cannot it just rely on the `expl_nodes` value?

adisos: The condition `(output_format == 'txt' or output_format == 'txt_no_fw_rules')` is actually the supported output formats for adding output explanation. I would add this as a method of `ExplTracker`.

adisos: Can the call to compute and add the explanation output be done earlier? (connectivity query?) it is specific to the connectivity query, and the `compute_final_results` is a general function for all queries, so I find it confusing that it handles explanations added for a specific query usecase.

adisos (changes requested):

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,425 @@
...
+from nca.CoreDS.ConnectivityProperties import ConnectivityProperties
+
+
+class ExplTracker(metaclass=Singleton):
+    """
+    The Explainability Tracker is used for tracking the elements and their configuration
+    so it will be able to specify which configurations are responsible for each peer and each connection
+    or lack of connection between them.
+
+    The ExplTracker is Singletone
```
adisos, line 20:
```suggestion
    The ExplTracker is Singleton
```

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,425 @@
...
+from nca.Utils.NcaLogger import NcaLogger
+from nca.CoreDS.Peer import PeerSet
+from bs4 import BeautifulSoup
+from bs4.element import Tag
+from nca.CoreDS.ConnectivityProperties import ConnectivityProperties
+
+
+class ExplTracker(metaclass=Singleton):
+    """
+    The Explainability Tracker is used for tracking the elements and their configuration
```
adisos, line 16:
```suggestion
    The Explainability Tracker is used for tracking the elements and their configuration,
```

adisos (commented):
output comments: for ip-block as `src` or `dst` - no need to add policy configurations or resources configurations.

adisos (commented):
Test run issue: 
when running with `explain: ALL` and without `-opt=true` , nca crashes in `explain_all` function, though I would expect this function should not be called.

nca/NetworkConfig/NetworkLayer.py
adisos, line 260: this can be added in the stage of adding the policy from `PoliciesFinder`, since optimized connectivity properties are already computed during parsing.

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,426 @@
...
+            all_conns - all the calculated connection. This is used to shortcut the check if 2 peers are connected or not
+
+            all_peers - all the peers. This is used for the explain-all feature.
+
+            ep - endpoints configurations (use either full_name for pod mode, or workload_name for deployments mode)
+    """
+
+    def __init__(self, ep=''):
+        self.ExplDescriptorContainer = {}
+        self.ExplPeerToPolicyContainer = {}
```
adisos, line 47:
```suggestion
        self.ExplDescriptorContainer = dict()  # a map from str (resource name) to a dict object with entries: 'path','line','workload_name'
        self.ExplPeerToPolicyContainer = dict() # a map from str (peer name) to ExplPolicies object
```

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,426 @@
...
+    class ExplPolicies:
+        """
+        ExplPolicies holds the policies affecting peers in relation to other peers.
+        That is, for each peer it holds all the peers in its egress and ingress and the policies
+        that has effect on the connection to that peers.
+        """
+
+        def __init__(self):
+            self.egress_dst = {}
+            self.ingress_src = {}
```
adisos, line 64: init as `dict()` and document in a comment as a `map from ... to ...`

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,426 @@
...
+            :param PeerSet egress_dst: the set of egress destinations peers to add the policy too
+            :param PeerSet ingress_src: the set of ingress source peers to add the policy too
+            """
+            self.all_policies.add(policy_name)
+
+            if egress_dst:
+                self._add_policy(egress_dst, self.egress_dst, policy_name)
+
+            if ingress_src:
+                self._add_policy(ingress_src, self.ingress_src, policy_name)
```
adisos, line 98: - confused by method names `_add_policy` vs `add_policy`
- confusing param names (e.g. `peer_list` would expect to be a `list` but is a `dict` + its documentation does not explain it's a map from ...  to .... )

nca/NetworkConfig/NetworkConfigQueryRunner.py
shmfr: > why do we need both active/inactive mode for `ExplTracker` and `expl_nodes` from output configuration to check before adding explanation in the output? cannot it just rely on the `expl_nodes` value?

This is for the cases when opt=false but --explain was given.



> The condition `(output_format == 'txt' or output_format == 'txt_no_fw_rules')` is actually the supported output formats for adding output explanation. I would add this as a method of `ExplTracker`.

Done. 



> Can the call to compute and add the explanation output be done earlier? (connectivity query?) it is specific to the connectivity query, and the `compute_final_results` is a general function for all queries, so I find it confusing that it handles explanations added for a specific query usecase.

Done.

nca/Utils/ExplTracker.py
shmfr: Done.

nca/Utils/ExplTracker.py
shmfr: Done.

shmfr:
> output comments: for ip-block as `src` or `dst` - no need to add policy configurations or resources configurations.

Done - "ipblocks had no configurations" explanation now replace the explanation for Ipblocks.


> Test run issue: when running with `explain: ALL` and without `-opt=true` , nca crashes in `explain_all` function, though I would expect this function should not be called.

I could not re-create the crash. Perhaps was tested on an older version? the current version protects this case.

nca/SchemeRunner.py
```diff
@@ -141,6 +142,9 @@ def run_scheme(self):
         global_ns_list = self._handle_resources_list(self.scheme.get('namespaceList', None))
         global_resource_list = self._handle_resources_list(self.scheme.get('resourceList', None))
         resources_handler = ResourcesHandler()
+        if self.optimized_run == 'true':
+            # we need to track configurations for the queries to use later-on
+            ExplTracker().activate()
```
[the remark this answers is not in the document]
shmfr: in nca_cli.py:

...
    if args.explain is not None and args.optimized_run == 'true':
        output_config['explain'] = args.explain
        ExplTracker().activate()
...

nca/NetworkConfig/TopologyObjectsFinder.py
```diff
@@ -217,6 +228,8 @@ def _add_hep_from_yaml(self, hep_object):
             hep.add_profile(profile)
 
         self._add_peer(hep)
+        if ExplTracker().is_active():
+            ExplTracker().add_item(hep_object.path, hep_object.line_number, hep.full_name())
```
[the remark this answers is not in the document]
shmfr: I am not sure I understand. The display names for the nodes are in the graph itself (svg for example). The explainability needs to describe the policies affecting them and where the configurations are. It can also get, as input, 2 nodes in the format of endpoints=deployments (with the type in Parenthesis).

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,426 @@
...
+                                     'Default-Policy',
+                                     egress_list,
+                                     ingress_list,
+                                     )
+
+    def is_policy_list_empty(self, node_name, check_ingress):
+        """
+        A service function to check if the expl' list of ingress or egress is empty.
+        :param node_name: the node to check
+        :param check_ingress: list to check (ingress or egress)
```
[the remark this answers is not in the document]
shmfr: Done.

nca/Utils/ExplTracker.py
shmfr: Done

nca/Utils/ExplTracker.py
shmfr: Done

nca/Utils/ExplTracker.py
shmfr: Done (changed the _add_policy name so it will be clearer. Also documented it better).

docs/SchemeFileFormat.md
```diff
@@ -86,7 +86,7 @@ The supported entries in the outputConfiguration object are as follows:
 | subset           | A dict object with the defined subset elements to display in the output                                         | [subset](#subset) object                                     |
 | fullExplanation  | Choose if to print all counterexamples causing the query result in the output                                   | bool                                                         |
 | excludeIPv6Range | If the policies of the config do not contain any IPv6 addresses, do not include IPv6 range in the query results | bool  [default: True]                                        |
-
+| explain          |A pair of node names to explain the policies affecting their connection or lack of it. When 'ALL' is supplied, all the pairs in the topology are explained.| string [ ns/node1,ns/node2 / ALL ]                           |
```
adisos, line 89: - should mention it is only supported for connectivity query.
- how it can be used for cli query? 
- add examples how this can be used, and example input/output (consider a separate section/page about this feature) .
- what if one of the nodes is an ip-block?

tests/k8s_cmdline_tests.yaml
```diff
@@ -422,3 +422,69 @@
     --pod_list helm_tests/helm_test6/
     --expected_output expected_cmdline_output_files/helm_test_resolved_yaml_next_to_multi_charts.txt
   expected: 0
+
+- name: basic_connectivity-expl
+  args: >
+    --connectivity k8s_testcases/example_policies/testcase8/testcase8-networkpolicy1.yaml
+    --ns_list k8s_testcases/example_podlist/ns_list.json
+    --pod_list k8s_testcases/example_podlist/pods_list.json
+    --explain ALL
```
adisos, line 431: missing a basic test with `explain` for two specific nodes

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
```diff
@@ -0,0 +1,14775 @@
+final fw rules for query: , config: testcase8-networkpolicy1.yaml:
+src: 0.0.0.0/0 dst_ns: [default,kube-system,kube-system-dummy-to-ignore,vendor-system] dst_pods: [*] conn: All connections
+src_ns: [default,kube-system,kube-system-dummy-to-ignore,vendor-system] src_pods: [*] dst: 0.0.0.0/0 conn: All connections
+src_ns: [default,kube-system,kube-system-dummy-to-ignore,vendor-system] src_pods: [*] dst_ns: [default,kube-system,kube-system-dummy-to-ignore,vendor-system] dst_pods: [*] conn: All connections
+<entry dst="0.0.0.0/0" id="1" src="0.0.0.0/0">
```
adisos, line 5: add a newline space before the explanations section start + a text header

adisos: I think it might be more useful to print the explanations in some textual format rather than xml (maybe consider this as another follow up issue to consider), especially when appended to the connectivity textual output

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
```diff
@@ -0,0 +1,14775 @@
...
+ <text>
+  Configurations affecting 0.0.0.0/0:
+IP blocks have no configurations
+ </text>
```
adisos, line 9: I think we can remove this entire section. it does not add any information.

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
```diff
@@ -0,0 +1,14775 @@
...
+</entry>
+<entry dst="default/cog-agents(DaemonSet)" id="2" src="0.0.0.0/0">
+ <text>
+  Configurations affecting the connectivity between (src)0.0.0.0/0 and (dst)default/cog-agents(DaemonSet):
+
+(src)0.0.0.0/0:
+IP blocks have no configurations
```
adisos, line 16: same here

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
```diff
@@ -0,0 +1,14775 @@
...
+
+(dst)default/cog-agents(DaemonSet):
+Policy Configurations:
+default/cog-agents(DaemonSet): line 5 in file ./k8s_testcases/example_podlist/pods_list.json
```
adisos, line 20: this is not the policy configuration, it's the node configuration

nca/NetworkConfig/TopologyObjectsFinder.py
```diff
@@ -194,6 +203,8 @@ def _add_networkset_from_yaml(self, networkset_object):
         for cidr in cidrs:
             ipb.add_cidr(cidr)
         self._add_peer(ipb)
+        if ExplTracker().is_active():
+            ExplTracker().add_item(networkset_object.path, networkset_object.line_number, ipb.full_name())
```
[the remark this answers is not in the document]
adisos: I think it may be redundant in the output (as well as html output) to specify anything about ip-blocks, even if the user may click on them.. (the explanation for connections between ip-block and workload may refer to policies involving only that workload).
@zivnevo  do you have any preference here?

nca/Utils/ExplTracker.py
```diff
@@ -0,0 +1,426 @@
...
+            self.ExplDescriptorContainer[full_name] = {'path': path, 'line': ln, 'workload_name': workload_name}
+        else:
+            NcaLogger().log_message('Explainability error: configuration-block name can not be empty', level='E')
+
+    def derive_item(self, new_name):
+        """
+        Handles resources that change their name after parsing, like virtual-service
+        that adds the service name and /allowed
+        :param str new_name: the name for the new derived element
+        """
```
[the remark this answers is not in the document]
adisos: the `param` line is now duplicated

nca/NetworkConfig/TopologyObjectsFinder.py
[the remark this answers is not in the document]
shmfr: OK, so a connection involving ipblocks does not have to include both src and dst configurations, but we look at it as a whole block of configurations. It may have 2 parts or one, but we look at it as one list of configurations explaining the connections. Thus, removing explanations for ipblocks.

nca/Utils/ExplTracker.py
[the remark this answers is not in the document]
shmfr: Done.

nca/NetworkConfig/NetworkLayer.py
shmfr: Done

docs/SchemeFileFormat.md
shmfr: Done. Also removed 'ALL' since it is for internal use.

tests/k8s_cmdline_tests.yaml
shmfr: Done

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
shmfr: After discussion, the 'ALL' option will be used for internal usage. The 2-nodes scenario already has a newline separation.

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
shmfr: Done

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
shmfr: Done

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
shmfr: Fixed.

docs/SchemeFileFormat.md
```diff
@@ -78,15 +78,15 @@ For example: `my_set/prod_ns/deny_all_policy`. If there are multiple policies na
...
+| fullExplanation  | Choose if to print all counterexamples causing the query result in the output                                                                                                                                                                                                  | bool                                                         |
+| excludeIPv6Range | If the policies of the config do not contain any IPv6 addresses, do not include IPv6 range in the query results                                                                                                                                                                | bool  [default: True]                                        |
+| explain          | A pair of node names to explain the policies affecting their connection or lack of it. Relevant only for connectivity query. Connections including IP-Blocks will show only the configurations of the node in that connection (since, IP-Blocks does not have configurations). | string [ ns/node1,ns/node2 ]                           |
```
adisos, line 89:
```suggestion
| explain          | A pair of node names (comma separated) to explain the policies affecting their connection or lack of it. Relevant only for connectivityMap query. Connections including IP-Blocks will show only the configurations of the node in that connection (since, IP-Blocks does not have configurations). | string [ ns/node1,ns/node2 ]                           |
```
In scheme file the query is named `connectivityMap`

README.md
```diff
@@ -101,6 +101,11 @@ The arguments to `--resource_list` and to `--base_resource_list` should be one o
 - `--output_endpoints`\
   Choose endpoints type in output (pods/deployments).\
   *default:* deployments
+  - `--explain`\
+  A pair of node names to explain the policies affecting their connection or lack of it. Relevant only for connectivity query.\
```
adisos, line 105:
```suggestion
  A pair of node names (comma separated) to explain the policies affecting their connection or lack of it. Relevant only for connectivity query.\
```

docs/SchemeFileFormat.md
adisos: so it is not possible to get `explain` for a connectivity between deployment and a certain ip-address? do we want to specify this, or create an issue to consider support this later?

tests/k8s_cmdline_tests.yaml
```diff
@@ -422,3 +422,80 @@
     --pod_list helm_tests/helm_test6/
     --expected_output expected_cmdline_output_files/helm_test_resolved_yaml_next_to_multi_charts.txt
   expected: 0
+
+- name: basic_connectivity_specific_nodes-expl
+  args: >
+    --connectivity k8s_testcases/example_policies/testcase8/testcase8-networkpolicy1.yaml
+    --ns_list k8s_testcases/example_podlist/ns_list.json
+    --pod_list k8s_testcases/example_podlist/pods_list.json
+    --explain default/cog-agents-d54st,default/cog-agents-qr8gp
```
adisos, line 431: in the documentation you specify args to explain as deploymnent names (such as `default/deployment-A1`), not pod names, but here you use pod names.. so which one is the correct format?

tests/expected_cmdline_output_files/basic_connectivity_specific_nodes_expl_output.txt
```diff
@@ -0,0 +1,18 @@
...
+Policy Configurations:
+Default-Policy
+Resource Configurations:
+default/cog-agents(DaemonSet): line 5 in file ./k8s_testcases/example_podlist/pods_list.json
+
+(dst)default/cog-agents(DaemonSet):
+Policy Configurations:
+allow-traffic-from-vendor-system-to-default-on-all-ports-typo: line 6 in file ./k8s_testcases/example_policies/testcase8/testcase8-networkpolicy1.yaml
+Resource Configurations:
+default/cog-agents(DaemonSet): line 67 in file ./k8s_testcases/example_podlist/pods_list.json
```
adisos, line 18: The output here is a bit awkward, as it shows the same name for src and dst, with different resource configurations

tests/expected_cmdline_output_files/basic_connectivity_expl_output.txt
adisos: I still see this section:
```
 <text>
  Configurations affecting 0.0.0.0/0:
 </text>
```

adisos: no need to create entry for `<entry dst="0.0.0.0/0" id="1" src="0.0.0.0/0">` , is both src and dst are ip- blocks it should be skipped.

tests/expected_cmdline_output_files/istio-ingress_expl_output.txt
```diff
@@ -0,0 +1,949 @@
...
+<entry dst="0.0.0.0/0" id="9" src="default/istio-ingressgateway(Deployment)">
+ <text>
+  Configurations affecting the LACK of connectivity between (src)default/istio-ingressgateway(Deployment) and (dst)0.0.0.0/0:
+
+(src)default/istio-ingressgateway(Deployment):
+Policy Configurations:
+Default-Policy
+None: line 1 in file ./istio_testcases/example_policies/istio-ingress-test/resources/virtual_services.yaml
+Resource Configurations:
+default/istio-ingressgateway(Deployment): line 1 in file ./istio_testcases/example_policies/istio-ingress-test/resources/istio-ingress-gateway-deployment.yaml
```
adisos, line 100: why is the policy configuration named `None`?

tests/k8s_cmdline_tests.yaml
shmfr: the user can use node names (both deployments or pods).

adisos: so this should also be specified in the doc.

docs/SchemeFileFormat.md
shmfr: Done

README.md
shmfr: Done

tests/k8s_cmdline_tests.yaml
shmfr: I mean, the type of the node is not relevant (that is why I used 'node'). I have diversified the example to include both a pod and a deployment to make it more clear.

tests/expected_cmdline_output_files/basic_connectivity_specific_nodes_expl_output.txt
shmfr: Fixed.

tests/expected_cmdline_output_files/istio-ingress_expl_output.txt
shmfr: Fixed.

adisos: Now `bookinfo-rule/eu.bookinfo.com, uk.bookinfo.com, reviews.prod.svc.cluster.local/allow:` is the policy name?

tests/expected_cmdline_output_files/basic_connectivity_specific_nodes_expl_output.txt
adisos: can you open an issue with details of the original example you were using here, explaining the issue and gaps between naming conventions of pods/depoloyments? also referring to the output of this explanations before the change.

docs/SchemeFileFormat.md
adisos: maybe you missed this comment from last iteration?

tests/expected_cmdline_output_files/basic_connectivity_specific_nodes_expl_output.txt
shmfr: Done.

docs/SchemeFileFormat.md
shmfr: explanation is possible for all nodes-pairs in the topology, including ip-blocks. The ip-blocks input format was added to the docs so it will be more clear how to use the feature.

adisos (approved):

PR #492, opened 2023-04-30
ingress policy parser optimization + add total runtime message

adisos:


tanyaveksler (approved):
In optimized implementation we should initialize `default_connsq` with `ConnectivityProperties.make_empty_props()`.

zivnevo (approved):

PR #493, opened 2023-04-30
Avoiding redundant warnings in VirtualService parsing

tanyaveksler:
Avoided redundant warnings (about fields that do not impact the analysis) in parsing of HTTPRroute in VirtualService.


nca/Parsers/IstioTrafficResourcesYamlParser.py
```diff
@@ -194,8 +194,8 @@ def parse_vs_http_route(self, vs, vs_spec):
         for route in http or []:
             self.check_fields_validity(route, f'HTTPRroute in the VirtualService {vs.full_name()}',
                                        {'name': [0, str], 'match': 0, 'route': 0, 'redirect': 3, 'delegate': 3,
-                                        'rewrite': 3, 'timeout': 3, 'retries': 3, 'fault': 3, 'mirror': 3,
-                                        'mirrorPercentage': 3, 'corsPolicy': 3, 'headers': 3})
+                                        'rewrite': 0, 'timeout': 0, 'retries': 0, 'fault': 0, 'mirror': 3,
+                                        'mirrorPercentage': 0, 'corsPolicy': 3, 'headers': 3})
```
adisos, line 198: I think the remaining fields with `3` can also be set to `0` instead.

tanyaveksler: The remaining fields might impact our analysis (if we support them)

adisos (approved):

PR #510, opened 2023-05-29
remove trivial conns from workload to itself in txt-no-fw-rules format

shireenf-ibm:
#519


adisos (changes requested):
hi @shireenf-ibm , looks good. 
Please create an issue to explain this change. and also update documentation for this output format.

adisos (changes requested):

nca/FWRules/ConnectivityGraph.py
```diff
@@ -304,33 +304,24 @@ def get_connections_without_fw_rules_txt_format(self, connectivity_msg=None):
             for connectivity-map output with connectivity restriction,
             or the type of connectivity changes for semantic-diff query output
         :rtype: str
-        :return: a string of the original peers connectivity graph content (without minimization of fw-rules)
+        :return: a string of the original peers connectivity graph content, excluding connections between workload to itself
```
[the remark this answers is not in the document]
shireenf-ibm: worth thinking about it -
-> on the one hand, I am not sure this keeps the correctness of the semantic diff result
but on the other hand,
how common is the cases of changes to self-loop connections for different configurations?!
(when having two network-policy sets: Is it common to change connections between workload to itself ?
or is it common to add/remove replicas of persistent peers when having two configuration topology?)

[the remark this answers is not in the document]
adisos: I think that for correctness considerations, we should consider such `connectivity-diff` in the output as well.

nca/FWRules/ConnectivityGraph.py
```diff
@@ -298,39 +298,31 @@ def _get_equals_groups(self):
...
-        # adding conns to itself for workloads with single replica
-        for workload_name in [wl for wl in workload_name_to_peers_map if len(workload_name_to_peers_map[wl]) == 1]:
-            lines.add(f'{workload_name} => {workload_name} : All Connections')
+                if src_peer != dst_peer:
+                    src_peer_name = self._get_peer_details(src_peer, True)[0]
+                    dst_peer_name = self._get_peer_details(dst_peer, True)[0]
+                    # no self-loops: if a peer has different replicas or copies, a connection from it to itself will
+                    # not be added either
+                    if exclude_self_loop_conns and src_peer_name == dst_peer_name:
+                        continue
```
adisos, line 322: can you also add a test for semantic-diff, for which a connection between workload to itself is in the output of the connectivity diff in this format?

adisos: ok, great! how did it not fail in one of the previous commits on which such connections were removed from the output? 
Is that example tested on txt_no_fw_rules format?  if not, let's add this format to that test as well.

adisos: now I noticed you've added this format to that test.. thanks!

adisos (approved):

PR #548, opened 2023-07-09
Interferes based queries optimized

tanyaveksler:


nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1895,6 +1901,20 @@ def exec(self, cmd_line_flag):
...
 
+    def check_interferes_optimized(self, cmd_line_flag=False):
+        conn_props1 = self.config1.allowed_connections_optimized()
+        conn_props2 = self.config2.allowed_connections_optimized()
+        conns1, conns2 = self.filter_conns_by_input_or_internal_constraints(conn_props1.allowed_conns,
+                                                                            conn_props2.allowed_conns)
+        if conns1.contained_in(conns2):
+            return QueryAnswer(False, self.name1 + ' does not interfere with ' + self.name2,
+                               numerical_result=0 if not cmd_line_flag else 1)
+
```
adisos, line 1912: how does this containment also considers `captured` as in the original version?

tanyaveksler: It considers `captured` connections because, as you can see a few lines above, `conns1 / conns2` are calculated based on `conn_props1.allowed_conns / conn_props2.allowed_conns` (these are the captured connections; all connections would be c`onn_props1.all_allowed_conns / conn_props2.all_allowed_conns`).

adisos (approved):

Issue #549, opened 2023-07-10
ingress policy parser issue


adisos:
Change from error to warning the case in the link below (missing port in the service), and just ignore that service rule in the analysis

https://github.com/IBM/network-config-analyzer/blob/master/nca/Parsers/IngressPolicyYamlParser.py#L97

zivnevo:
Also, see if the message can provide more information. I would expect something like
```
Ingress rule redirects traffic to <service-name>:<port-number>, but port <port-number> is not exposed by <service-name>.
```

PR #551, opened 2023-07-10
fix issue #549 + add test

adisos:


nca/Parsers/IngressPolicyYamlParser.py
```diff
@@ -94,7 +94,11 @@ def parse_backend(self, backend, is_default=False):
 
         service_port = srv.get_port_by_name(port_name) if port_name else srv.get_port_by_number(port_number)
         if not service_port:
-            self.syntax_error(f'Missing port {port_name if port_name else port_number} in the service', service)
+            port_str = f'{port_name if port_name else port_number}'
+            warning_msg = f'Ingress rule redirects traffic to {service_name}:{port_str}, '
+            warning_msg += f' but port {port_str} is not exposed by Service {service_name}'
+            self.warning(warning_msg, service)
```
zivnevo, line 100:
```suggestion
            self.warning(warning_msg, backend)
```
Wouldn't it be better to point the user to the problematic Ingress rule?

zivnevo (approved):

PR #556, opened 2023-07-11
Intersects based queries optimized

tanyaveksler:


nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1932,13 +1932,22 @@ def exec(self, cmd_line_flag):
 class IntersectsQuery(TwoNetworkConfigsQuery):
     """
     Checking whether both configs allow the same connection between any pair of peers
+    Note: this query is only used by ForbidsQuery.
+    It's not symmetrical: config1 is the one to be checked, and only its captured connections are considered,
+    while config2 is a dummy one (that defines things to be checked in config1), and all its connections are considered.
```
adisos, line 1937: isn't this the opposite? for the config to be checked - all its connections should be considered, and for the dummy config - only its captured connections are considered.

tanyaveksler: I don't think so. dummy config might have nothing captured.
I wrote the comment according to the code:
```
                conns1 = conns1_captured if only_captured else conns1_all
                conns2, _, _, _ = self.config2.allowed_connections(peer1, peer2)
```
Note, the `only_captured `flag is true (in `ForbidsQuery`, which is the only usage of `IntersectsQuery`.

adisos: taking a look at the `forbids` description from readme: `--forbids - Checking whether the base set of NetworkPolicies forbids the traffic explicitly specified in the given set of NetworkPolicies`.
Isn't the `explicitly` means only captured? and the policies to check is the base set?

adisos: @zivnevo can you add your input here?

zivnevo: I recall switching the roles of the base config and the forbids config at some point. I'm not sure which one is used for which role now. Maybe @shireenf-ibm remembers more.

Anyway, one of the configs is used to specify the connections the user want to ensure are forbidden in the other config. Only the connections that are **explicitly** allowed in the "specification" config are being checked in the "implementation" config. That is, only connections for which at least one of their endpoint is captured, so that "allowed by default" connections are not considered. The reason for this confusing specification is that K8s has no deny actions (we later tried to avoid this confusion by letting the user use our "baseline rules" language).

adisos (approved):

PR #576, opened 2023-07-26
simplify_output flag

haim-kermany:
added a flag to avoid putting nodes in a namespace square in a dot file.
needed for cases when graphwiz has issues with the created dot file


nca/FWRules/DotGraph.py
```diff
@@ -74,9 +76,11 @@ def add_node(self, subgraph, name, node_type, label):
         param label: node label
         """
         label = [tok.strip() for tok in label if tok != '']
+        title = subgraph if self.do_not_subgraph else ''
+        subgraph = '' if self.do_not_subgraph else subgraph
         if subgraph not in self.subgraphs:
             self.subgraphs[subgraph] = self.Subgraph(subgraph)
-        node = self.Node(name, node_type, label)
+        node = self.Node(name, node_type, label, title)
```
adisos, line 83: lines 81-82 should not check if `subgraph` is empty  string and ignore it in that case?

nca/nca_cli.py
```diff
@@ -326,6 +327,9 @@ def nca_main(argv=None):
     parser.add_argument('--file_out', '-f', type=str, help='A file path to which output is redirected')
     parser.add_argument('--expected_output', type=str, help='A file path of the expected query output,'
                                                             'relevant only with --connectivity and --semantic_diff')
+    parser.add_argument('--simplify_output', action='store_true',
+                        help='simplify the connectivity graph,'
+                             'relevant only when output_format is dot or jpg')
```
adisos, line 332: - consider renaming the flag `simplify_output` to be more specific, such as `simplify_graph`
- add documentation about the new flag option on `md` files (main readme that lists all flgas, and scheme file options that details outputConfiguration options).

adisos (commented):
Added a few comments.
The graph generated on the example from the issue does look much better, though I still get the same `graphviz` error about trouble in init_rank.

nca/FWRules/DotGraph.py
haim-kermany: no,
there is  also a subgraph without name . 
and all nodes w/o a subgraph are at this subgraph
it is also later given as a parameter to _subgraph_to_str()

nca/nca_cli.py
haim-kermany: done

haim-kermany: done

adisos (approved):

PR #604, opened 2023-09-19
preparation to support istio egress gateway

tanyaveksler:


nca/NetworkConfig/ResourcesHandler.py
```diff
@@ -69,24 +69,24 @@ def get_full_livesim_resource_path(livesim_resource_path):
...
 
     @staticmethod
-    def get_relevant_livesim_resources_paths_by_labels_matching(livesim_resource_path, missing_resource_labels_dict):
+    def get_relevant_livesim_resources_paths_by_labels_matching(livesim_resource_path, missing_resource_labels):
         """
         check by labels matching if one of the livesim resources has matching labels for a resource referenced by one
         of the parsed policies. If yes, return its path to be added to the configuration, to enable the analysis.
         :param str livesim_resource_path: a path to the relevant livesim dir to check for resources
-        :param dict missing_resource_labels_dict: the labels from parsed policy in the config for
+        :param list missing_resource_labels: the labels from parsed policy in the config for
```
adisos, line 77: isn't `missing_resource_labels` a set of pairs rather than a list?

nca/NetworkConfig/ResourcesHandler.py
```diff
@@ -69,24 +69,24 @@ def get_full_livesim_resource_path(livesim_resource_path):
...
         resource_full_path = ResourcesHandler.get_full_livesim_resource_path(livesim_resource_path)
         livesim_resource_labels = ResourcesParser.parse_livesim_yamls(resource_full_path)
-        for key in missing_resource_labels_dict.keys():
+        for (key, value) in missing_resource_labels:
             for yaml_path, labels in livesim_resource_labels.items():
-                if missing_resource_labels_dict.get(key) == labels.get(key):
+                if (key, value) in labels:
                     res.append(yaml_path)
-        return res
+        return list(dict.fromkeys(res))  # remove duplicates
```
adisos, line 89: instead of having to remove duplicates, `res` can be a `set()`,  and the return value `list(res)`

nca/NetworkConfig/ResourcesHandler.py
tanyaveksler: Right. Updated.

nca/NetworkConfig/ResourcesHandler.py
tanyaveksler: Done

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -422,36 +421,30 @@ def _allowed_xgress_conns_optimized(self, is_ingress, peer_container, res_conns_
...
+        if captured_res:
+            all_allowed_conns = allowed_conns
         return PolicyConnections(captured=captured_res, allowed_conns=allowed_conns, denied_conns=ConnectionSet(),
                                  all_allowed_conns=all_allowed_conns)
 
     def _allowed_xgress_conns_optimized(self, is_ingress, peer_container, res_conns_filter=PolicyConnectionsFilter()):
-        res_conns = OptimizedPolicyConnections()
-        all_peers_and_ips = peer_container.get_all_peers_group(True)
-        all_peers_no_ips = peer_container.get_all_peers_group()
+        # TODO - should we add non-captured connections from mesh to dns_entries??
```
adisos, line 432: Do you have an example or can you explain this comment?

nca/NetworkConfig/NetworkLayer.py
```diff
@@ -422,36 +421,30 @@ def _allowed_xgress_conns_optimized(self, is_ingress, peer_container, res_conns_
...
+        all_peers_and_ips = peer_container.get_all_peers_group(add_external_ips=True, include_dns_entries=True)
```
adisos, line 433: Why do you include `dns_entries` here but not in non-optimized solution?

nca/Resources/IngressPolicy.py
```diff
@@ -39,8 +40,8 @@ def contained_in(self, other):
 class IngressPolicy(NetworkPolicy):
     """
     This class implements ingress controller logic for incoming http(s) requests
-    The logic is kept similarly to NetworkPolicy, where the selected_peers are the ingress controller peers,
-    and the rules are egress_rules.
+    The logic is kept similarly to NetworkPolicy, where the selected_peers are the ingress/egress controller peers,
+    and the rules are ingress/egress_rules.
```
adisos, line 44: Should we consider renaming this class? `IngressPolicy` is no longer for ingress gateway, but also for egress gateway.

nca/NetworkConfig/NetworkLayer.py
tanyaveksler: Removed this comment. Instead, we agreed to add deny rules for connections from mesh to DNS entries

nca/NetworkConfig/NetworkLayer.py
tanyaveksler: In non-optimized solution: `all_allowed_conns = ConnectionSet(True)` in case of non-captured, which is equivalent to the above optimized solution.

nca/Resources/IngressPolicy.py
adisos: I see it used also to represent k8s Ingress policy, so maybe remove the `Istio` from the name?
Also, documentation of this class should specify the relevant resources (Istio Ingress/Egress gateway and k8s Ingress )

nca/Resources/GatewayPolicy.py
```diff
@@ -29,18 +30,18 @@ def __eq__(self, other):
...
 
 
-class IngressPolicy(NetworkPolicy):
+class GatewayPolicy(NetworkPolicy):
     """
     This class implements ingress controller logic for incoming http(s) requests
-    The logic is kept similarly to NetworkPolicy, where the selected_peers are the ingress controller peers,
-    and the rules are egress_rules.
+    The logic is kept similarly to NetworkPolicy, where the selected_peers are the ingress/egress controller peers,
+    and the rules are ingress/egress_rules.
```
adisos, line 44:
```suggestion
    The logic is kept similarly to NetworkPolicy, where the selected_peers are the ingress/egress controller peers,
    and the rules are ingress/egress_rules.
    This class is used to represent policies from `k8s Ingress` , `istio IngressGateway` and `istio EgresGateway`
```

nca/NetworkConfig/NetworkConfigQuery.py
```diff
@@ -1282,7 +1282,7 @@ def clone_without_ingress(config):
             return config  # no ingress policies in this config
         config_without_ingress = config.clone_without_policies(config.name)
         for policy in config.policies_container.policies.values():
-            if not isinstance(policy, IngressPolicy):  # ignoring ingress policies
+            if not isinstance(policy, GatewayPolicy):  # ignoring ingress policies
```
adisos, line 1285:
```suggestion
            if not isinstance(policy, GatewayPolicy):  # ignoring gateway policies
```

adisos, line 1285: consider renaming the function `clone_without_ingress`, as now the layer and resources are not named `ingress` but rather `gateway`

tanyaveksler: Done

adisos (approved):

PR #619, opened 2023-10-10
fix code scanning alert

shireenf-ibm:


zivnevo (approved):
Looks good.
Thanks for handling this mess. I like it better without the "look behind" operator.

PR #676, opened 2024-01-28
Istio gateway policies explainability information

tanyaveksler:
Added origin path/line information to istio gateway policies, for explainability report.


haim-kermany (approved):
the code works for me
thanks.
BTW, do we still need to used derive_item() for the sidecar?

haim-kermany (approved):

adisos (approved):

PR #678, opened 2024-01-29
handle multi quary with html

haim-kermany:


nca/SchemeRunner.py
```diff
@@ -183,6 +181,40 @@ def get_query_output_config_obj(self, query):
         output_config_obj = OutputConfiguration(output_configuration_dict, query['name'])
         return output_config_obj
 
+    def cativate_expTracker(self, query_array):
```
adisos, line 184: rename func to `activate_exp_tracker`

nca/SchemeRunner.py
```diff
@@ -183,6 +181,40 @@ def get_query_output_config_obj(self, query):
...
+        """
+        check if it is safe to activate the ExplTracker, and activate it
+        activating is safe if we have at most one query that needs explainabilty, and it must be the first
+        :param list[dict] query_array: A list of query objects to run
+        :return: weather it safe to run the queries
```
adisos, line 189:
```suggestion
        :return: whether it safe to run the queries
```

nca/SchemeRunner.py
```diff
@@ -183,6 +181,40 @@ def get_query_output_config_obj(self, query):
         output_config_obj = OutputConfiguration(output_configuration_dict, query['name'])
         return output_config_obj
 
+    def acativate_exp_tracker(self, query_array):
```
adisos, line 184: there is still a typo here to fix..  `activate` instead of  `acativate`

PR #717, opened 2024-04-16
Fix package name

zivnevo:


zivnevo:
> How did work until now, with `nca_build/network-config-analyzer-*.tar.gz` ?

Seems like there was a recent change in the way Python packages are named when running `python -m build`.
I'm not quite sure how this change propagated to our pipeline. I'll continue to investigate

